PR #3, opened 2018-03-13
Add template file organization

teamdisc:
I added groups and files following our discussion regarding our standard file organization. So we will stick with MVC architecture first, then what should be changed will be up to each project requirement. As @olivierobert suggested, we could also have multiple template projects for each type as well. The file organization for our current template project will look like this:

screen shot 2018-03-13 at 5 32 21 pm


teamdisc:
closing this one due to integration of `Bakery` script requiring support for project name to be included with spacing. more info in PR #5

Issue #14, opened 2018-03-28
Integrate SwiftLint


teamdisc:
Integrate SwiftLint to the project using Cocoapods and pre-define set of rules that will be required for our template code conventions.

PR #17, opened 2018-03-28
Integrate SwiftLint

teamdisc:
## SwiftLint integration

I worked on integrating **SwiftLint** into our template project. I have dived into [Rules](https://github.com/realm/SwiftLint/blob/master/Rules.md) and made some configurations for our `.swiftlint.yml` file. 

There're a couple of rules I opted in and out. Also, Some of the rules I might wanna discuss with you guys whether to include them or not. 🤔

### disabled rules
1. **trailing_whitespace**

### opt-in rules
1. **attributes** - Attributes should be on their own lines in functions and types, but on the same line as variables and imports.
2. **closure_spacing** - Closure expressions should have a single space inside each brace.
3. **contains_over_first_not_nil** - Prefer `contains` over `first(where:) != nil`
4. **empty_count** - Prefer checking `isEmpty` over comparing `count` to zero.
5. **empty_string** - Prefer checking `isEmpty` over comparing `string` to an empty string literal.
6. **explicit_init** - Explicitly calling ``.init()`` should be avoided.
7. **extension_access_modifier** - Prefer to use extension access modifiers
8. **fatal_error_message** - A `fatalError` call should have a message.
9. **first_where** - Prefer using `.first(where:)` over `.filter { }.first` in collections.
10. **implicitly_unwrapped_optional** - Implicitly unwrapped optionals should be avoided when possible.
11. **joined_default_parameter** - Discouraged explicit usage of the default separator.
12. **literal_expression_end_indentation** - Array and dictionary literal end should have the same indentation as the line that started it.
13. **multiline_arguments** - Arguments should be either on the same line, or one per line.
14. **multiline_parameters** - Functions and methods parameters should be either on the same line, or one per line.
15. **nimble_operator** - Prefer Nimble operator overloads over free matcher functions.
16. **operator_usage_whitespace** - Operators should be surrounded by a single whitespace when they are being used.
17. **overridden_super_call** - Some overridden methods should always call super
18. **pattern_matching_keywords** - Combine multiple pattern matching bindings by moving keywords out of tuples.
19. **private_action** - IBActions should be private.
20. **prohibited_super_call** - Some methods should not call super
21. **redundant_nil_coalescing** - nil coalescing operator is only evaluated if the lhs is nil, coalescing operator with nil as rhs is redundant
22. **single_test_class** - Test files should contain a single QuickSpec or XCTestCase class.
23. **sorted_first_last** - Prefer using `min()` or `max()` over `sorted().first` or `sorted().last`
24. **trailing_closure** - Trailing closure syntax should be used whenever possible.
25. **unneeded_parentheses_in_closure_argument** - Parentheses are not needed when declaring closure arguments.
26. **untyped_error_in_catch** - Catch statements should not declare error variables without type casting.
27. **vertical_parameter_alignment_on_call** - Function parameters should be aligned vertically if they're in multiple lines in a method call.
28. **yoda_condition** - The variable should be placed on the left, the constant on the right of a comparison operator.

### configured rules
- **line_length** - to ignore function declarations & comments

### discussion-needed rules
- **closure_end_indentation** - Closure end should have the same indentation as the line that started it.
- **discouraged_optional_boolean** - Prefer non-optional booleans over optional booleans.
- **discouraged_optional_collection** - Prefer empty collection over optional collection.
- **explicit_type_interface** - Properties should have a type interface
- **file_length** - need configured?
- **force_unwrapping** - should we enable this one? seems required, but well, it could cause a lot of annoyances.
- **large_tuple** - should make some configuration?
- **object_literal** - should we go for literal object over normal init?
- **override_in_extension** - Extensions shouldn't override declarations.
- **prefixed_toplevel_constant** - Top-level constants should be prefixed by `k`. should we enable this one? IMO, I personally feel `k` is a legacy way and pretty old-school 🙈. But I like the idea of creating Constants struct. Feel more Swifty to me.
- **private_outlet** - If we enable `private_action`, do we have to enable this one as well? Normally, we'd make these `IBOutlet`s private. But in some certain case, I think general views can leaked theirs `IBOutlet`s sometimes. Say it's a common view that doesn't contain any logic, but only its responsibility is just to have other view configure it. 🤔
- **quick_discouraged_call** - Discouraged call inside 'describe' and/or 'context' block. Seems we should enable it, but I don't quite understand it. anyone?
- **quick_discouraged_focused_test** - Discouraged focused test. Other tests won't run while this one is focused.
- **quick_discouraged_pending_test** - Discouraged pending test. This test won't run while it's marked as pending. Need to check, if it's just a warning, then I think we should enable it. (think we can configure its danger level anyways)

I just learned that making unit test function `private` is to skip that test case. hmm 🤔


ProudOfZiggy:
17. **overridden_super_call** - Some overridden methods should always call super

I'm not sure this is a good idea because for some functions from default SDK you don't need to call `super`. Especially functions of default UIView drawing cycle.
For example `UIView.draw(_:)` `super` doesn't make any sense because it's doing nothing. Also, in some cases you might want to do something like this:
`
class AbstractClass {
    func doSomething() {
        fatalError(message: "Implementation of abstract class must be overridden in its subclass")
    }
}

class ConcreteClass {
    override func doSomething() {
        //Concrete Implementation
    }
}
`

25. **unneeded_parentheses_in_closure_argument** - Parentheses are not needed when declaring closure arguments.

This rule is actually good. BUT. In my practice there was a situation when Swift compiler didn't want to work w/o Parentheses in case of complicated closure. I don't have an example because it was some time ago, but this can happen. But I like this rule.

26. **untyped_error_in_catch** - Catch statements should not declare error variables without type casting.

We might add it, yes. But, historically, we are not using a lot of try/catch in iOS development. So is this a point for this?

27. **vertical_parameter_alignment_on_call** - Function parameters should be aligned vertically if they're in multiple lines in a method call.

Isn't it conflicting or how it's working with **multiline_arguments**? Just curious.

ProudOfZiggy:
**discussion-needed rules**

**closure_end_indentation**

I agree with that rule because of good indentation code might me much readable. I'm voting for that rule.

**discouraged_optional_boolean**

Also agree. Because I've seen some people using `Optional Bool` and its can be really annoying to work with. Because, I mean, it's a boolean, it should have two possible conditions, not three. I'm voting for that rule.

**discouraged_optional_collection**

Same thing here. I'm voting for that rule.

**explicit_type_interface**

I don't think wee need this rule, because swift gave us a beautiful opportunity to avoid doing this. For me, by RHS you can easily understand what type is this.

**file_length**

I don't think we need this rule, because even if we a trying to do controllers very thin there could be a files that reasonably can be vey large. A very good example of such file is a some cool custom UI component that could contain a huge amount of custom drawing using Core Graphics and other stuff. It easy for this files to be 2000+ rows of code.

**force_unwrapping**

This is a good rule and I also don't like force unwrapping. I'm voting for that rule.

**large_tuple**

We can add it, but for me it's actually a very rare situation when someone wanna create a bug tuple.

**object_literal**

This one is tricky. I prefer to use `Object Literal`. BUT. They could lead to situation like this:
screen shot 2018-03-29 at 18 12 02
Soooo, I will vote for this actually. But we still better to use Files with constants for colors and stuff like this.

**override_in_extension**

I don't think we need this rule. At some point those overriding could be bad because of its implicitly. BUT. But they could be a good "weapon" when you are using protocol oriented programming a lot, they might help you get more flexibility at some point.

ProudOfZiggy:
**discussion-needed rules**

**prefixed_toplevel_constant**

You are absolutely right. This is old style. In Swift is a bad practice to use such constants. You need to incapsulate everything. For me only some basic (like mathematic functions) can be on top level.

**private_outlet**

We no need this rule I think. Because there could be situation where you have a UITableViewCell with only 1 or 2 @IBOutlet and its ok to configure their content directly from controller.

ProudOfZiggy:
**trailing_whitespace**

Why to disable this? =)

teamdisc:
Thanks for all the comments man. all great points! @ProudOfZiggy  ✌🏻

> **overridden_super_call** - I'm not sure this is a good idea because for some functions from default SDK you don't need to call super. Especially functions of default UIView drawing cycle.
For example UIView.draw(_:) super doesn't make any sense because it's doing nothing.

Aha, only some of the functions will be required to call `super` by SwiftLint tho. But it's a good thing that you pointed that out. so I've dived into their codebase and found this one out:
```swift
private let defaultIncluded = [
        //NSObject
        "awakeFromNib()",
        "prepareForInterfaceBuilder()",
        //UICollectionViewLayout
        "invalidateLayout()",
        "invalidateLayout(with:)",
        "invalidateLayoutWithContext(_:)",
        //UIView
        "prepareForReuse()",
        "updateConstraints()",
        //UIViewController
        "addChildViewController(_:)",
        "decodeRestorableState(with:)",
        "decodeRestorableStateWithCoder(_:)",
        "didReceiveMemoryWarning()",
        "encodeRestorableState(with:)",
        "encodeRestorableStateWithCoder(_:)",
        "removeFromParentViewController()",
        "setEditing(_:animated:)",
        "transition(from:to:duration:options:animations:completion:)",
        "transitionCoordinator()",
        "transitionFromViewController(_:toViewController:duration:options:animations:completion:)",
        "viewDidAppear(_:)",
        "viewDidDisappear(_:)",
        "viewDidLoad()",
        "viewWillAppear(_:)",
        "viewWillDisappear(_:)",
        //XCTestCase
        "setUp()",
        "tearDown()"
    ]
```

---------------------------------

> **unneeded_parentheses_in_closure_argument**  - This rule is actually good. BUT. In my practice there was a situation when Swift compiler didn't want to work w/o Parentheses in case of complicated closure. I don't have an example because it was some time ago, but this can happen. But I like this rule.

I faced some compiler reporting a weird error about this parentheses thing also. but in my case, it was my fault. the compiler just reported an error that wasn't related to the actual cause. I think we can try having it enabled first and see how it goes from there. If it doesn't play well with the compiler, then we can disable it later. what do you think? 🤔

---------------------------------

> **untyped_error_in_catch** - We might add it, yes. But, historically, we are not using a lot of try/catch in iOS development. So is this a point for this?

I do agree on that part, but IMO, having this enabled surely wouldn't hurt us even if we just use it occasionally. But it will only help us when needed. So I'm in for having this one on. what do you think?

---------------------------------

> **vertical_parameter_alignment_on_call** - Isn't it conflicting or how it's working with **multiline_arguments**? Just curious.

Actually, I think it will support each other. To combine these rules, it will only accept things like:
```swift
func someFunction(param1: String,
                  param2: String,
                  param3: String)
```
the former rule will only allow `param1`, `param2`, and `param3`  to be aligned to each other. The latter rule will only allow all the params to stay on different lines (when we need it not to be inline).

teamdisc:
> **closure_end_indentation** - I agree with that rule because of good indentation code might me much readable. I'm voting for that rule.

I'd love to have it on also, but let's see if our guys are all on the same page here. @jasonnam @Dekablade01 @olivierobert 

--------------------------

> **discouraged_optional_boolean** - Also agree. Because I've seen some people using Optional Bool and its can be really annoying to work with. Because, I mean, it's a boolean, it should have two possible conditions, not three. I'm voting for that rule.

Yeah, I agree on this one as well. Because it makes more sense for boolean to only have two choices, but my concern was that there were other cases that I might've missed. Good point! ✌🏻

--------------------------

> **discouraged_optional_collection** - Same thing here. I'm voting for that rule.

Yup, we're cool on this one then.

--------------------------

> **explicit_type_interface** - I don't think wee need this rule, because swift gave us a beautiful opportunity to avoid doing this. For me, by RHS you can easily understand what type is this.

I'm not sure about this one also. Personally, I'd not like to enable it 😂 But our guys might wanna have it on. So I brought it into discussion. Anyway, I'd explicitly declare type interface when it comes to `CGFloat`, `Float`, and `Double`. but I'm concerned that it could introduce inconsistency to the project. what do you guys think? 🤔

--------------------------

> **file_length** - I don't think we need this rule, because even if we a trying to do controllers very thin there could be a files that reasonably can be vey large. A very good example of such file is a some cool custom UI component that could contain a huge amount of custom drawing using Core Graphics and other stuff. It easy for this files to be 2000+ rows of code.

Good point! I agree, but we could also disable any rules with one comment.
```swift
// swiftlint:disable file_length
```

--------------------------

> **force_unwrapping** - This is a good rule and I also don't like force unwrapping. I'm voting for that rule.

I strongly agree on this one and I want this rule enabled also. but my concern is that there are cases that we could perhaps require to declare some object with force unwrapping. These objects are mostly related to the view itself. Let's say if there's a class that requires injected on initialization with some certain view and that view will only be created after `viewDidLoad` is called (`IBOutlets`). 

So that's why I'd like to discuss with you guys first to see if we should enable it or not, but IMO, I'm in for this rule.

--------------------------

> **large_tuple** - We can add it, but for me it's actually a very rare situation when someone wanna create a bug tuple.

lol we should go for it then.

--------------------------

> **object_literal** - This one is tricky. I prefer to use Object Literal. BUT. They could lead to situation like this. Soooo, I will vote for this actually. But we still better to use Files with constants for colors and stuff like this.

 Sure, we will still use constants for colors and stuff. I think we will use literal object mostly on `UIImage`. For colors if it's to be used, it will only be used in Color constants file anyway I guess.

--------------------------

> **override_in_extension** - I don't think we need this rule. At some point those overriding could be bad because of its implicitly. BUT. But they could be a good "weapon" when you are using protocol oriented programming a lot, they might help you get more flexibility at some point.

I'm actually voting for this one tho. I agree on the part that it could be used a lot and handy to do so, but we can explicitly disable this rule when required sometimes. I'm afraid that some hidden implementation might cause confusion for other team members when they just jumped into the project and expect some class to perform its expected behavior normally.

This is very much open for discussion. what do you think tho? I'm still 50/50 about it, but think I'll vote for this one 🙈

ProudOfZiggy:
> I'm not sure about this one also. Personally, I'd not like to enable it 😂 But our guys might wanna have it on. So I brought it into discussion. Anyway, I'd explicitly declare type interface when it comes to CGFloat, Float, and Double. but I'm concerned that it could introduce inconsistency to the project. what do you guys think?

Well, I'm voting for NOT adding it. And for Double you ca just write `var a = 0.0` =)

> I faced some compiler reporting a weird error about this parentheses thing also. but in my case, it was my fault. the compiler just reported an error that wasn't related to the actual cause. I think we can try having it enabled first and see how it goes from there. If it doesn't play well with the compiler, then we can disable it later. what do you think? 🤔

We can try I think.

> I do agree on that part, but IMO, having this enabled surely wouldn't hurt us even if we just use it occasionally. But it will only help us when needed. So I'm in for having this one on. what do you think?

Also can try.

> I strongly agree on this one and I want this rule enabled also. but my concern is that there are cases that we could perhaps require to declare some object with force unwrapping. These objects are mostly related to the view itself. Let's say if there's a class that requires injected on initialization with some certain view and that view will only be created after viewDidLoad is called (IBOutlets).

As we discussed, this rule will not be triggered by such situation. So we can, and I think we should, use it.

Dekablade01:
close #14

PR #26, opened 2018-03-30
Setup Template with Code Signing

Dekablade01:
1. Moved `CocoaPods` to `Bundle`
2. Install Fastlane in `Bundle`
3. Init Fastlane with `Appfile` & `Gymfile` for building and releasing to Appstore
4. Code Signing, this will download or create `distribution` and `development` certificates and create or `.mobileprovision` for each build config. 

For debug setting use `development` provision, `staging` which we're going to upload to `Fabric` use `ad-hoc` provision and for appstore use `app-store` provision.

will code #19


__PROJECT NAME__/fastlane/.env
```diff
@@ -0,0 +1,44 @@
...
+SCHEME_NAME="__PROJECT NAME__"
+SCHEME_NAME_STAGING="${SCHEME_NAME} staging"
+SCHEME_NAME_DEBUG="${SCHEME_NAME} debug"
+
+################
+## SLACK
+################
+# SLACK_URL=
+# SLACK_USERNAME=
+# SLACK_ICON_URL=
```
teamdisc: let's leave some comment here and don't forget to update it in our `README.md` ✌🏻

__PROJECT NAME__/fastlane/Fastfile
```diff
@@ -0,0 +1,94 @@
...
+    # AppStore Mobile Provisioning ( RELEASE )
+    sigh(
+      username: "#{ENV['APPLE_ID']}",
+      app_identifier: "#{ENV['BUNDLE_ID']}",
+      output_path: "./code_signing/profiles/appstore",
+    )
+  end
+
+  desc "setup_template"
+  lane :setup_template do
```
teamdisc: how about changing its name to be more specific? perhaps something like `setup_app_on_developer_portal` and let's put in some comment on how it works

PR #48, opened 2018-08-15
Create AppID helper lane

Dekablade01:
### What's Happened ###
- After setup the project, then we need the app id for code signing 

## Insight ### 
- These lanes will create AppID on `Developer portal` with project name that set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,10 @@ ROOT
   - Integrate UI and unit test
 
 #### Code Signing
+  - AppID
+    - AppIDs can be created automatically with `lane :create_app_ids`, this will create AppIDs for `Appstore`, `Staging` and `Debug` schemes.
+  - App Store Connect 
+    - App in App Store Connect needs to be created manually. 
   - Certificate
     - Development
       - for DEBUG build configuration.
--- a/__PROJECT NAME__/fastlane/Developer Portal/Codefile
+++ b/__PROJECT NAME__/fastlane/Developer Portal/Codefile
@@ -80,21 +80,27 @@ platform :ios do
   end
 
   desc "create App ID for all app bundles"
-  lane :create_app_ids do
+  lane :create_app_ids do |options|
     ## DEBUG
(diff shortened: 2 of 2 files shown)


__PROJECT NAME__/fastlane/Developer Portal/Codefile
```diff
@@ -78,4 +78,37 @@ platform :ios do
...
+  desc "create App ID for all app bundles."
+  private_lane :create_app_on_developer_portal do |options|
+    produce(
+      username: "#{ENV["APPLE_ID"]}",
+      app_identifier: options[:app_identifier],
+      app_name: options[:app_name].gsub("-", " "),
+      language: options[:language] || "English",
+      app_version: options[:app_version] || "1.0",
+      sku: options[:sku],
+      team_name: options[:team_name],
```
teamdisc: correct if I'm wrong, but from what I understand is that all the options here will actually be ignored, right?

If so, should we add `options` as a parameter to `create_app_ids` and pass it along to this function as well? 🤔

README.md
```diff
@@ -41,6 +41,10 @@ ROOT
   - Integrate UI and unit test
 
 #### Code Signing
+  - AppID
+    - AppIDs can be created automatically with `lane :create_app_ids`, this will create AppIDs for `Appstore`, `Staging` and `Debug` schemes.
+  - App Store Connect 
+    - App in App Store Connect need to be created manually.
```
teamdisc: `needs` ✌🏻

__PROJECT NAME__/fastlane/Developer Portal/Codefile
Dekablade01: Yes, some options can be ignore. 

'create_app_ids' lane will be executed manually with fastlane cli. It's possible to make optional options for this lane. But it' will be hard to execute with option because we have to pass these parameters with cli and as I see parameters of produce, there's nothing we need to custom.

Issue #160, opened 2021-08-11
Setup schemes


phongvhd93:

blyscuit:
Issues blocking this PR
- #165

PR #167, opened 2021-08-19
[#160] Setup schemes with Tuist

llleyelll:
Resolves https://github.com/nimblehq/ios-templates/issues/160

## What happened

Setup schemes:
- \<ProjectName\>
- \<ProjectName\> Staging
 
## Insight

- Each property's`configurationName` should be updated after implement build configurations.
 
## Proof Of Work

Screen Shot 2021-08-19 at 4 10 55 PM


Tuist/ProjectDescriptionHelpers/Scheme+Initializing.swift
```diff
@@ -0,0 +1,36 @@
...
+            shared: true,
+            buildAction: BuildAction(targets: ["\(name)"]),
+            testAction: TestAction(
+                targets: ["\(name)Tests", "\(name)UITests"],
+                configurationName: "Debug"
+            ),
+            runAction: RunAction(configurationName: "Debug"),
+            archiveAction: ArchiveAction(configurationName: "Release"),
+            profileAction: ProfileAction(configurationName: "Debug"),
+            analyzeAction: AnalyzeAction(configurationName: "Debug")
```
suho, line 17: Please add `#warning` or comments for these configuration names, we should use `Debug Production` and `Release Production`.

Issue #166, opened 2021-08-18
Generating the project follows the standard file structure


vnntsu:
## Why

Following the [standard file structure](https://www.notion.so/nimblehq/iOS-Templates-ac74dffa740744be9ac2d5c3905856dc?p=63c00b90a81a4a41b1ae3df73dbf2738) document.

With Tuist, we're able to generate a template with that standard

```json
.
├── README.md
├── {ProjectName}
│   ├── Configurations
│   │   ├── Plists
│   │   └── XCConfigs
│   ├── Resources
│   │   ├── Assets
│   │   ├── Languages
│   │   └── LaunchScreen
│   └── Sources
│       ├── Application
│       │   ├── AppDelegate.swift
│       │   ├── Application.swift
│       │   └── SceneDelegate.swift
│       ├── Constants
│       │   ├── Constants+API.swift
│       │   └── Constants.swift
│       ├── Data
│       │   ├── Keychain
│       │   │   ├── Keychain.swift
│       │   │   ├── KeychainKey.swift
│       │   │   └── Models
│       │   ├── NetworkAPI
│       │   │   ├── AuthenticatedNetworkAPI.swift
│       │   │   ├── NetworkAPI.swift
│       │   │   ├── Core
│       │   │   ├── Interceptors
│       │   │   ├── Models
│       │   │   └── RequestConfigurations
│       │   └── Repositories
│       │       ├── RepositoryProvider.swift
│       │       ├── Authentication
│       │       └── User
│       ├── Domain
│       │   ├── Entities
│       │   │   ├── User.swift
│       │   │   └── Token.swift
│       │   ├── Interfaces
│       │   │   └── Repositories
│       │   └── UseCases
│       │       ├── UseCaseProvider.swift
│       │       ├── Authentication
│       │       └── User
│       ├── Presentation
│       │   ├── Modules
│       │   │   ├── Home
│       │   │   └── Login
│       │   ├── Navigator
│       │   │   ├── Navigator+Scene.swift
│       │   │   ├── Navigator+Transition.swift
│       │   │   └── Navigator.swift
│       │   └── Views
│       │       ├── Button
│       │       ├── CollectionView
│       │       ├── TextField
│       │       └── Transition
│       └── Supports
│           ├── Builder
│           │   └── Builder.swift
│           ├── Extensions
│           │   ├── Foundation
│           │   ├── Rx
│           │   └── UIKit
│           └── Helpers
│               ├── Rx
│               ├── Typealias
[...]

PR #177, opened 2021-08-31
[166] Generating: Source folder

vnntsu:
https://github.com/nimblehq/ios-templates/issues/166

## What happened

Generating follow the structure:

```json
│       ├── Data
│       │   ├── NetworkAPI
│       │   │   ├── AuthenticatedNetworkAPI.swift
│       │   │   ├── NetworkAPI.swift
│       │   │   ├── Core
│       │   │   ├── Interceptors
│       │   │   ├── Models
│       │   │   └── RequestConfigurations
│       │   └── Repositories
│       ├── Domain
│       │   ├── Entities
│       │   ├── Interfaces
│       │   └── UseCases
│       │       ├── UseCaseProvider.swift
│       ├── Presentation
│       │   ├── Modules
│       │   ├── Navigator
│       │   │   ├── Navigator+Scene.swift
│       │   │   ├── Navigator+Transition.swift
│       │   │   └── Navigator.swift
│       │   └── Views
```
 
## Insight

- Prepare some empty folder with `.gitkeep` inside
- Target is now including empty folders as the `resources`. I will find a script to remove it after generating the project.
- Steal the `Core` service from `mvvm-rxswift-demo` repo, and some protocols.

## Proof Of Work

Screen Shot 2021-08-31 at 13 33 29

### DIFF
--- a/ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
+++ b/ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
@@ -40,7 +40,6 @@ extension NetworkAPIProtocol {
                 return Disposables.create()
             }
         }
-        .observe(on: MainScheduler.instance)
         .asSingle()
     }
 }
--- a/ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
+++ b/ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
@@ -27,7 +27,6 @@ extension NetworkAPIProtocol {
             interceptor: configuration.interceptor
         )
         .responseData()
-        .observe(on: ConcurrentDispatchQueueScheduler(qos: .background))
         .flatMap { _, data -> Observable<T> in
             Observable.create { observer in
                 do {
--- a/ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
+++ b/ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
@@ -1,0 +1,46 @@
+//
+//  NetworkAPIProtocol.swift
+//
+
+import RxSwift
+import Alamofire
+import RxAlamofire
+
+protocol NetworkAPIProtocol {
+
+    func performRequest<T: Decodable>(_ configuration: RequestConfiguration, for type: T.Type) -> Single<T>
+}
+
+extension NetworkAPIProtocol {
+
+    func request<T: Decodable>(
+        session: Session,
+        configuration: RequestConfiguration,
+        decoder: JSONDecoder
+    ) -> Single<T> {
+        return session.rx.request(
+            configuration.method,
+            configuration.url,
+            parameters: configuration.parameters,
+            encoding: configuration.encoding,
+            headers: configuration.headers,
+            interceptor: configuration.interceptor
+        )
+        .responseData()
+        .observe(on: ConcurrentDispatchQueueScheduler(qos: .background))
+        .flatMap { _, data -> Observable<T> in
+            Observable.create { observer in
+                do {
+                    let decodable = try decoder.decode(T.self, from: data)
+                    observer.on(.next(decodable))
+                } catch {
+                    observer.on(.error(error))
+                }
+                observer.on(.completed)
+                return Disposables.create()
+            }
+        }
+        .observe(on: MainScheduler.instance)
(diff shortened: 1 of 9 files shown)


ProjectName/Sources/Data/NetworkAPI/NetworkAPI.swift
```diff
@@ -0,0 +1,23 @@
...
+
+import Foundation
+import Alamofire
+import RxSwift
+
+final class NetworkAPI: NetworkAPIProtocol {
+
+    private let decoder: JSONDecoder
+
+    init(decoder: JSONDecoder = JSONAPIDecoder.default) {
```
suho, line 13: We didn't add `JSONMapper` lib into this template, so I think we just need to use `JSONDecoder`. Please define the default instance also.
```suggestion
    init(decoder: JSONDecoder = JSONDecoder.default) {
```

ProjectName/Sources/Presentation/Navigator/Navigator.swift
```diff
@@ -0,0 +1,76 @@
...
+    func show(scene: Navigator.Scene, sender: UIViewController?, transition: Navigator.Transition)
+    func dismiss(sender: UIViewController?)
+    func pop(sender: UIViewController?)
+    func popToRoot(sender: UIViewController?)
+}
+
+final class Navigator {
+
+    private func viewController(from scene: Scene) -> UIViewController {
+        // Make view controller
```
suho, line 19: We should make the project "buildable" after generating the project, so I suggest:
```suggestion
        #warning("Create view controllers here")
        return HomeViewController()
```

ProjectName/Sources/Presentation/Navigator/Navigator.swift
```diff
@@ -0,0 +1,76 @@
...
+        }
+        guard let sender = sender else {
+            fatalError("Need sender for navigation or modal transition")
+        }
+
+        switch transition {
+        case .navigation(let interactionTransition):
+            if var target = target as? AdoptedInteractiveTransition {
+                target.transition = interactionTransition
+            }
```
suho, line 53: We should remove these lines

ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
```diff
@@ -0,0 +1,46 @@
...
+                    let decodable = try decoder.decode(T.self, from: data)
+                    observer.on(.next(decodable))
+                } catch {
+                    observer.on(.error(error))
+                }
+                observer.on(.completed)
+                return Disposables.create()
+            }
+        }
+        .observe(on: MainScheduler.instance)
```
phongvhd93, line 43: Can you give more insight into this, why we block the UI thread when calling an API?

This `.observe(on: MainScheduler.instance)` is not flexible to me because sometimes we might call another API after the first call.

`firstAPICall.flatMap { secondAPICall() } .subscribe()`

minhnimble: should we bring the observe part into the usecase? this way we can control when should we observe (whether it is a chain request or not). What do you guys think? 🤔

suho, line 43: Agree with @phongvhd93 and @minhnimble, I think we can remove `observe` operators here, and the developer can control more by bringing the observe part into the usecase
```suggestion
        .flatMap { _, data -> Observable<T> in
            Observable.create { observer in
                do {
                    let decodable = try decoder.decode(T.self, from: data)
                    observer.on(.next(decodable))
                } catch {
                    observer.on(.error(error))
                }
                observer.on(.completed)
                return Disposables.create()
            }
        }
```

ProjectName/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
```diff
@@ -0,0 +1,45 @@
...
+        return session.rx.request(
+            configuration.method,
+            configuration.url,
+            parameters: configuration.parameters,
+            encoding: configuration.encoding,
+            headers: configuration.headers,
+            interceptor: configuration.interceptor
+        )
+        .responseData()
+        .observe(on: ConcurrentDispatchQueueScheduler(qos: .background))
```
suho, line 30: I think we can remove this line also, we will control about `observer(on:)` in use case is better

suho (approved):

Issue #132, opened 2021-02-04
Add swiftlint rules following our code convention


markgravity:
## Feature request

Add  swiftlint rules to enforce to follow our code convention which defines in Compass
https://nimblehq.co/compass/development/code-conventions/swift/

suho:
@markgravity Any update on this ticket?

markgravity:
@suho Let me plan to finish this task in next week 🙏

PR #207, opened 2021-10-13
[#132] Add swiftlint rules following our code convention

markgravity:
https://github.com/nimblehq/ios-templates/issues/132

## What happened

- Add swiftlint rules following our code convention
- Create base swiftlint rule file
 
## Insight
_Please help me with name of the rule and message, it's not perfect enough 🙏_ 

**Added custom rules**
- Double Decimal Point
```swift
// Bad
let width: CGFloat = 320

 // Good
let width: CGFloat = 320.0
```

- Declaration Vertical whitespace
```swift
// Bad
class User {
    let name: String
    let age: Int
}

// Good
class User {

    let name: String
    let age: Int
}
```

- Newline Between Switch Case
```swift
// Bad
switch value {
case .caseA:
    print("A")
    doA()

case .caseB:
    print("B")
    doB()

case .caseC:
    print("C")
    doC()

}

// Good
switch value {
case .caseA:
    print("A")
    doA()
case .caseB:
    print("B")
    doB()
case .caseC:
    print("C")
    doC()
}
```

## Proof Of Work

Screen Shot 2021-10-13 at 15 39 51

Screen Shot 2021-10-13 at 15 39 46


suho:
Screen Shot 2021-10-14 at 12 42 49

@markgravity Why only 1 warning on case 2? how about case 1 and case 3?

Issue #209, opened 2021-10-13
Clear all warnings on fresh generated project


markgravity:
## Why

With a fresh generated project, it shouldn't contain any warnings
 
Currently
<img width="434" alt="Screen Shot 2021-10-13 at 16 38 27" src="https://user-images.githubusercontent.com/17875522/137108319-5541a971-f21f-4139-b5ef-6d87fd885adb.png">

## Acceptance Criteria
- [ ] No warnings from Xcode after build in fresh generated project

suho:
@markgravity Please help me to implement this chore ticket

PR #232, opened 2021-11-24
[#209] Clear all warnings on fresh generated project

markgravity:
https://github.com/nimblehq/ios-templates/issues/209

## What happened

- Clear xcode warnings
- Update Rswift 6.0 build script follow [migration guide](https://github.com/mac-cain13/R.swift/blob/master/Documentation/Migration.md)
 
## Insight

n/a
 
## Proof Of Work
Screen Shot 2021-11-24 at 11 01 33


suho:
@markgravity Why you disable `unusedArguments`, `hoistPatternLet` rules?

markgravity:
> @markgravity Why you disable `unusedArguments`, `hoistPatternLet` rules?
@suho 
For  [unusedArgument](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#unusedArguments)`, it will warn every time you create a function with parameters but you don't use all of them. 
So in my option, it's unnecessary it better leave the name of parameters for later use 😁 

```swift
func application(
   _ application: UIApplication, // warning
   didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? // warning
) -> Bool {}

// FIXED
func application(
   _ _: UIApplication,
   didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]? 
) -> Bool {}
```

For [hoistPatternLet](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistPatternLet)
```
// This is the way we usually do right?
- if case .foo(let bar, let baz) = quux {
    // inner foo
  }

+ if case let .foo(bar, baz) = quux {
    // inner foo
  }
```

suho:
@markgravity 
About `unusedArguments`, I prefer to keep it and use `--stripunusedargs closure-only`.

With `hoistPatternLet`, we are using both ways, and almost everyone agrees with the rule, so don't need to remove it 😁
Now we should use:
```swift
if case let .foo(bar, baz) = quux {
    // inner foo
}
```

markgravity:
> @markgravity About `unusedArguments`, I prefer to keep it and use `--stripunusedargs closure-only`.
> 
> With `hoistPatternLet`, we are using both ways, and almost everyone agrees with the rule, so don't need to remove it 😁 Now we should use:
> 
> ```swift
> if case let .foo(bar, baz) = quux {
>     // inner foo
> }
> ```

Updated them 😁

suho (approved):

llleyelll (approved):

Issue #254, opened 2022-03-10
Show warnings for long compile code


markgravity:
## Why

_**Lesser build time -> Faster development**_

So that we should add these flags into `Build Settings > Other Swift Flags` to show warnings for long compile code
```
-Xfrontend -warn-long-function-bodies=300
-Xfrontend -warn-long-expression-type-checking=300
```
![Screen Shot 2022-03-10 at 10 39 10](https://user-images.githubusercontent.com/17875522/157589253-f0b569c5-d22a-4b08-8298-9136ed4378e1.png)

Refer the [discussion](https://nimble-co.slack.com/archives/C418DBL2W/p1646883596899949)
 
## Who Benefits?

iOS Developers

PR #256, opened 2022-03-22
[#254] Show warnings for long compile code

markgravity:
https://github.com/nimblehq/ios-templates/issues/254

## What happened

- Add warning flags into `Other Swift Flags` add to xconfig files: Debug Staging & Debug Production
 
## Insight

We decided to set the warning time value is `300` in this [RFC](https://github.com/nimblehq/ios-templates/issues/258)
 
## Proof Of Work

Screen Shot 2022-04-12 at 14 29 16

_Reduce the value to `1` to produce the warnings_
Screen Shot 2022-04-15 at 13 29 41


.github/workflows/test_install_script.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       run: bundle install
 
     - name: Start Install Script for Template App
-      run: sh make.sh --bundle-id co.nimblehq.template --bundle-id-staging co.nimblehq.template.staging --project-name TemplateApp
+      run: sh make.sh --bundle-id co.nimblehq.template --bundle-id-staging co.nimblehq.template.staging --project-name TemplateApp --compile-time-warning 200
```
suho, line 25: I suggest creating an RFC and deciding a default `warning-compile-time` instead of input here.

markgravity: @suho it's only for test, we can choose whatever value even leave it empty to skip it

suho: I mean that we don't need to init a value `warning-compile-time` from the setup script, we can set it directly with `OTHER_SWIFT_FLAGS`, for example:

```swift
public func createConfiguration(projectName: String,) -> Configuration {
	let settings: [String: SettingValue] = [
		"OTHER_SWIFT_FLAGS": "-Xfrontend -warn-long-expression-type-checking=200 -Xfrontend -warn-long-function-bodies=200"
	]

	let xcconfig = Path("\(projectName)/\(path)")
	switch self {
	case .debugStaging, .debugProduction:
		return .debug(name: name, settings: settings, xcconfig: xcconfig)
	case .releaseStaging, .releaseProduction:
		return .release(name: name, settings: settings, xcconfig: xcconfig)
	}
}
```

We don't need to always use the script setup to set some value like compile code time. If we all agree with `adding compile code time` into the project, then add it directly.

Tuist/ProjectDescriptionHelpers/BuildConfiguration.swift
```diff
@@ -30,13 +30,18 @@ public enum BuildConfiguration: CaseIterable {
         }
     }
 
-    public func createConfiguration(projectName: String) -> Configuration {
+    public func createConfiguration(projectName: String, compileTimeWarning: Int?) -> Configuration {
+        var settings: [String: SettingValue] = [:]
+				if let compileTimeWarning = compileTimeWarning {
+						settings["OTHER_SWIFT_FLAGS"] = "-Xfrontend -warn-long-expression-type-checking=\(compileTimeWarning) -Xfrontend -warn-long-function-bodies=\(compileTimeWarning)";
+				}
```
suho, line 37: Please check the indent here.
I would prefer you set up: `Prefer Indent Using: Spaces` instead of Tabs :D

make.sh
```diff
@@ -85,6 +96,15 @@ if ! [[ $bundle_id_production =~ $regex ]]; then
     die "Invalid Package Name: $bundle_id_production (needs to follow standard pattern {com.example.package})"
 fi
 
+if [ -z "$compile_time_warning" ] ; then
+    compile_time_warning="nil"
+else
+		regex='^(.[0-9]*)$'
+		if ! [[ $compile_time_warning =~ $regex ]]; then
+	    	die "Invalid Compile Time value: $compile_time_warning (needs to be integer)"
+		fi
```
suho, line 105: Please check the indent in this file also :D

.github/workflows/test_install_script.yml
suho: @markgravity Moreover, can we add `OTHER_SWIFT_FLAGS` into these [xcconfig files](https://github.com/nimblehq/ios-templates/tree/master/%7BPROJECT_NAME%7D/Configurations/XCConfigs) instead?

suho:
@markgravity It would be great to merge this pull request before end of this month 🥳

make.sh
```diff
@@ -20,21 +20,24 @@ usage() {
...
 Options:
--h, --help                                   display this usage message and exit
--b, --bundle-id [BUNDLE_ID_PRODUCTION]       the production id (i.e. com.example.package)
--s, --bundle-id-staging [BUNDLE_ID_STAGING]  the staging id (i.e. com.example.package.staging)
--n, --project-name [PROJECT_NAME]            the project name (i.e. MyApp)
+-h, --help                                            display this usage message and exit
+-b, --bundle-id [BUNDLE_ID_PRODUCTION]                the production id (i.e. com.example.package)
+-s, --bundle-id-staging [BUNDLE_ID_STAGING]           the staging id (i.e. com.example.package.staging)
+-n, --project-name [PROJECT_NAME]                     the project name (i.e. MyApp)
+-c, --compile-time-warning [COMPILE_TIME_WARNING]     the compile time warning (i.e. 200,300)
```
blyscuit, line 27: I think we should add units for clarity, this will be hard to remember in the long run.
```suggestion
-c, --compile-time-warning [COMPILE_TIME_WARNING]     the compile time warning in milliseconds (i.e. 200,300)
```

blyscuit (approved):
Looks good, some comments.

suho:
@markgravity Don't forget to change to base branch to `develop` (not master)

suho (approved):

minhnimble (approved):

suho:
@markgravity For proof of work, could you show me a function that takes more than 300ms when compiling? Just want to make sure there is a warning message? :D

markgravity:
@suho updated the POW

suho:
@markgravity Where is the POW screenshot? :D

markgravity:
> @markgravity Where is the POW screenshot? :D

Oops, I forgot hit `Update` 😓

Issue #275, opened 2022-05-09
Fix: Package name contains "-" symbol is invalid while running install script


ducbm051291:
## Issue

When run install script of ios-template using package name contains "-" symbol. Script failed to run with invalid package name

<img width="571" alt="Screen Shot 2022-05-09 at 13 49 31" src="https://user-images.githubusercontent.com/24598204/167355900-b5ada966-9f3c-4c41-8058-bdeea8be1b54.png">
 
## Expected

Allow package name contain "-" symbol and script should be able to run without Invalid package name error
 
## Steps to reproduce

1. Create repository using ios-template
2. Run install script with bundle have "-" symbol

PR #282, opened 2022-05-24
[#275] Fix: Package name contains "-" symbol is invalid while running install script

ducbm051291:
https://github.com/nimblehq/ios-templates/issues/275

## What happened

Install script currently does not accept "-" symbol inside the package name. We must update regex to allow that case.
 
## Insight

Update regex for package name validation
 
## Proof Of Work

Screen Shot 2022-05-24 at 15 11 28


markgravity (approved):
LGTM 🎉

suho (approved):
LGTM

minhnimble (approved):

Issue #291, opened 2022-06-10
Create Github Wiki document to introduce Fastlane set up in this template


nmint8m:
## Why

Currently, there is no document for Fastlane in this template. Besides, in this template, Fastlane pipelines are involved in the CI platform (such as Github Action, Bitrise). So it is important to have a technical documentation introducing Fastlane's functionalities, so as to save time during project setup and maintenance.
 
## Who Benefits?

- Who uses this template
- The base for issue #247

PR #294, opened 2022-06-13
[#291] [Chore] Create Github Wiki document to introduce Fastlane set up in this template

nmint8m:
https://github.com/nimblehq/ios-templates/issues/291

## What happened

Currently, there is no document for Fastlane in this template. Besides, in this template, Fastlane pipelines are involved in the CI platform (such as Github Action, Bitrise). So it is important to have a technical documentation introducing Fastlane's functionalities, so as to save time during project setup and maintenance.
 
## Insight

I created Github Wiki document to introduce Fastlane set up in this template.
 
## Proof Of Work

N/A


.github/wiki/Fastlane.md
```diff
@@ -0,0 +1,161 @@
+# About
```
suho, line 1: I think we can remove this line, as we already have a title "Fastlane", so it's okay to have a description as an introduction after the title.

.github/wiki/Fastlane.md
```diff
@@ -0,0 +1,161 @@
...
+
+Fastlane is a part of the automation tool for the development and release process of the project. This document will mention some crucial parts of the pipeline.
+
+<img src="assets/images/infrastructure/fastlane/fastlane.png" width=300>
+
+# Fastfile
```
suho, line 7: Let's change the level to `h2` ~ `##` instead, as the title of this document is `h1` already. Otherwise, the font size of this section will be the same with the title
```md
## Fastfile

### Build and test

### Synchronize the provisioning profile

...
## Matchfile
```

suho (approved):

vnntsu (approved):

Issue #293, opened 2022-06-13
Fix: There is no lane name `run_xcov` in Fastfile


nmint8m:
## Issue

Currently, the lane `run_xcov` does not exist in `Fastfile`. It was removed in PR #242.
 
## Expected

We should remove where the lane `run_xcov` is used, such as:

### In `bitrise.yml`

```yml
  test:
    steps:
    - ...
    - fastlane@3:
        inputs:
        - lane: run_xcov
```

### In `Bitrise.md`

```
## Workflows and Steps
...
| Fastlane - Run XCov | Fastlane Match | Fastlane Match | Fastlane Match |
```

PR #242, opened 2021-12-06
[#225] Improve Code Coverage Report in Dangerfile

suho:
Resolves #225

## What happened

Currently, in a pull request, we are showing the code coverage of all files in the project, we need to update Dangerfile to show only code coverage of file changes.
 
## Insight

- Filter file changes in pull requests.
- Show code coverage of all file changes instead of all files of the project.
 
## Proof Of Work

- No code file changes
Screen Shot 2021-12-06 at 14 07 42

- Have file changes:
Screen Shot 2021-12-06 at 14 07 52

PR #295, opened 2022-06-13
[#293] [Bug] Fix: There is no lane name run_xcov in Fastfile

nmint8m:
https://github.com/nimblehq/ios-templates/issues/293

## What happened

Currently, the lane run_xcov does not exist in Fastfile. It was removed in PR https://github.com/nimblehq/ios-templates/pull/242. So we should remove where the lane run_xcov is used.
 
## Insight

In this PR, I removed where the lane `run_xcov` is used in `bitrise.yml` and `Bitrise.md`.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -156,9 +156,6 @@ workflows:
     - fastlane@3:
         inputs:
         - lane: build_and_test
-    - fastlane@3:
-        inputs:
-        - lane: run_xcov
     - danger@2:
         inputs:
         - github_api_token: "$DANGER_GITHUB_API_TOKEN"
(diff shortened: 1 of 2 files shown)


bitrise.yml
```diff
@@ -154,9 +154,6 @@ workflows:
     - fastlane@3:
         inputs:
         - lane: build_and_test
-    - fastlane@3:
-        inputs:
-        - lane: run_xcov
```
suho, line 159: Based on [this step](https://github.com/nimblehq/ios-templates/blob/master/.github/workflows/automatic_pull_request_review.yml#L50) from the GitHub action workflow, please run `clean_up_xcov` instead and then update the Bitrise document also.

blyscuit (approved):

bitrise.yml
nmint8m: @suho I created the issue #296 to cover your comment. Please take a look.

suho (approved):

Issue #286, opened 2022-06-01
Fix: The trigger setup for the publish docs to wiki workflow


suho:
## Issue

The publish docs to wiki workflow is triggered on push with the path `docs/**`, however, we changed the path to `.github/wiki`, we need to update that
 
## Expected

- After release a new version, the publish docs to wiki workflow should be triggered.
 
## Steps to reproduce

1. Push code to the master
2. The publish docs to wiki workflow will not run

PR #299, opened 2022-06-14
[#286] [Bug] Fix: The trigger setup for the publish docs to wiki workflow

nmint8m:
https://github.com/nimblehq/ios-templates/issues/286

## What happened

The publish docs to wiki workflow is triggered on push with the path docs/**, however, we changed the path to .github/wiki, we need to update that
 
## Proof Of Work

N/A


suho (approved):
LGTM

vnntsu (approved):

Issue #301, opened 2022-06-30
Fix: Constants.DEV_PORTAL_APPLE_ID is unused by mistake


markgravity:
## Issue

Currently, `Constants.DEV_PORTAL_APPLE_ID` is required but it's not used anywhere

Refer this [discussion](https://nimble-co.slack.com/archives/C418DBL2W/p1656562397054919)

## Expected

`Constants.DEV_PORTAL_APPLE_ID` should be used in `MatchManager.rb` as `username`
 
## Steps to reproduce

n/a

PR #326, opened 2022-08-19
[#301] [Fix] Constants.DEV_PORTAL_APPLE_ID is unused by mistake

markgravity:
https://github.com/nimblehq/ios-templates/issues/301

## What happened

- Pass `DeliverableConstants.DEV_PORTAL_APPLE_ID` to `MatchManager` and `register_new_device` lane
 
## Insight

n/a
 
## Proof Of Work

Screen Shot 2022-08-19 at 13 48 04

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -34,7 +34,8 @@ match_manager = MatchManager.new(
   fastlane: self,
   keychain_name: Constants.KEYCHAIN_NAME,
   keychain_password: Constants.KEYCHAIN_PASSWORD,
-  is_ci: Environments.CI
+  is_ci: Environments.CI,
+  username: DeliverableConstants.DEV_PORTAL_APPLE_ID
 )
 
 before_all do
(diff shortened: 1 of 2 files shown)


blyscuit (commented):
Shouldn't the `username` be in `else` block and not in `is_ci` block? For `is_ci` we should not sync with App Store Connect. We sync with App Store connect only when local.

suho:
@markgravity cc @blyscuit Another concern, do we need to have a username as we are using AppStore Connect API Key?

markgravity:
> Shouldn't the `username` be in `else` block and not in `is_ci` block? For `is_ci` we should not sync with App Store Connect. We sync with App Store connect only when local.

Good catch, it should only include in local only 😓

blyscuit:
> @markgravity cc @blyscuit Another concern, do we need to have a username as we are using AppStore Connect API Key?

We could still use it for local run when we don't use API Key 🤔 . Maybe we can change the description text in `Deliverable_Constants.rb` to explain this more or we can remove it to use `matchfile` locally instead.

markgravity:
> > @markgravity cc @blyscuit Another concern, do we need to have a username as we are using AppStore Connect API Key?
> 
> We could still use it for local run when we don't use API Key 🤔 . Maybe we can change the description text in `Deliverable_Constants.rb` to explain this more or we can remove it to use `matchfile` locally instead.

Just wonder why we don't use the API key in local? 🤔

blyscuit:
> 
> @blyscuit Just wonder why we don't use the API key in local? 🤔

In my opinion, on local, you should be able to choose either way that fits you 🤔 Using username/password is more simple for a not often run command like match.

blyscuit (approved):

phongvhd93 (approved):
Look good

Shayokh144 (approved):
all look ok

blyscuit (approved):

Issue #321, opened 2022-08-17
Test Workflow upload the project to TestFlight


blyscuit:
## Why

Currently, we test the template only on building. Testing uploading the build to TestFlight would validate our template usability in a real world scenario.

## Acceptance Criteria

- [x] Create a new project under Nimble's account.
- [x] Create match profiles under Nimble iOS Github.
- [x] Add App Store Connect credential to Github Actions Secret. 
- [x] Create a script to modify Constants to match Secrets.
> APP_STORE_KEY_ID
> APP_STORE_ISSUER_ID
> APPSTORE_CONNECT_API_KEY
> Match Repository
- [x] Add a new temporary fastlane script to upload build to TestFlight, if needed.
- [x] Modify [test_install_script.yml](https://github.com/nimblehq/ios-templates/blob/master/.github/workflows/test_install_script.yml) workflow to perform the upload.
- [x] Remove temporary scripts in `make.sh` remove step.

## Who Benefits?

iOS developers.

blyscuit:
Currently having issue with fastlane https://github.com/fastlane/fastlane/issues/20553

blyscuit:
@markgravity This is done. Please close when finalize reviewed.

blyscuit:
@markgravity Can we close this one?

PR #329, opened 2022-08-23
[#321] Test Workflow upload the project to TestFlight

blyscuit:
#321 

## What happened

Test uploading the build to TestFlight for each commits.
 
## Insight

- Update SSH-Agent version on Github Actions.
- Use base64 string for App Store API key to reduce error from newline secret.
- Add document about base64 for usage instruction.
- Add `.github/workflows/test_testflight.yml` to run new test.
- Add `set_up_test_testflight.sh` for configuring TestFlight ready app.
- Modify `make.sh` to remove test TestFlight flow files.
- Add `SKIP_FIREBASE_DSYM` ENV for project with no Firebase analytics.
- Fix `BuildManager.rb` using global var instead of local var.
- Remove `DEV_PORTAL_TEAM_ID` because it's replaceable by `match` ENV.

## Proof Of Work

Screen Shot 2022-08-23 at 10 32 55
Screen Shot 2022-08-23 at 09 33 23


fastlane/Fastfile
```diff
@@ -198,7 +204,7 @@ platform :ios do
     app_store_connect_api_key(
       key_id: DeliverableConstants.APP_STORE_KEY_ID,
       issuer_id: DeliverableConstants.APP_STORE_ISSUER_ID,
-      key_content: Environments.APPSTORE_CONNECT_API_KEY || ''
+      key_filepath: "appstorekey.p8"
```
phongvhd93, line 207: Can you give more insight about using the .p8 file instead of `APPSTORE_CONNECT_API_KEY`?

ducbm051291: @phongvhd93   Regarding Apple's document about security, we shouldn't store keys in our repository so I think @blyscuit using `p8` file is reasonable

Screen Shot 2022-08-29 at 09 01 48

blyscuit: @phongvhd93  Before we use `APPSTORE_CONNECT_API_KEY` from CI secret, which is already very safe.
Currently fastlane has this bug https://github.com/fastlane/fastlane/issues/20553 that cannot read `APPSTORE_CONNECT_API_KEY` as environment secret.
In this PR we save `APPSTORE_CONNECT_API_KEY` from secret to `.p8` only on the CI machine. This is safe for GH Actions Runner but not safe for self-hosted machine, so we add an extra step to remove the file when the job is done. Value from this .p8 cannot be print, see in POW.
When this bug is fixed by fastlane we can remove these p8 related steps.

blyscuit: We NEVER push p8 or APPSTORE_CONNECT_API_KEY directly on the repo.

phongvhd93 (approved):
All good

.github/workflows/deploy_AppStore.yml
```diff
@@ -69,6 +69,13 @@ jobs:
       env:
         MATCH_PASSWORD: ${{ secrets.MATCH_PASS }}
 
+    # Fix issue with https://github.com/fastlane/fastlane/issues/20553.
+    - name: Set Up App Store API Secret  
+      env:
+        APP_STORE_KEY: ${{ secrets.APPSTORE_CONNECT_API_KEY }}
+      run: |
+        echo "$APP_STORE_KEY" > appstorekey.p8
```
suho, line 77: Based on [this practice](https://docs.github.com/en/actions/security-guides/encrypted-secrets#limits-for-secrets), I would prefer to use `gpg` and `the AES256 cipher algorithm` or `base64`. What do you think?

.github/workflows/test_testflight.yml
```diff
@@ -0,0 +1,80 @@
+name: Test Upload Build to TestFlight
```
suho, line 1: We can improve the name: `test_upload_build_to_test_flight.yml`

.github/workflows/deploy_AppStore.yml
suho: @blyscuit Another way to fix this is using `distribution_manager.upload_to_testflight` instead. It will use `fastlane.pilot` instead of `fastlane.deliver`

.github/workflows/test_testflight.yml
```diff
@@ -0,0 +1,80 @@
...
+### SSH_PRIVATE_KEY for Match Repo
+### MATCH_REPO
+### MATCH_PASS
+### API_KEY_ID
+### ISSUER_ID
+### APPSTORE_CONNECT_API_KEY
+
+on:
+  push:
+    branches: [ feature/**, bug/**, chore/** ]
```
suho, line 13: I would prefer to test this on the `develop` branch only, otherwise, we will spam a lot of builds on the TestFlight page. What do you think?

blyscuit: @suho Should we use on:pull_request instead? The idea is to run this test to check if the code submitted for PR can still build successfully, if it run on develop then the failed code might already be merge to develop.

suho: Yes, just think about this again, we could do with `on:pull_request` :D

.github/workflows/deploy_AppStore.yml
blyscuit: The current issue if with `fastlane. app_store_connect_api_key`. This PR might be best to wait until the error is fixed. I will retry with a different approach for GH Secrets but since `fastlane.app_store_connect_api_key` didn't work locally, I have some concern. 
The practices outlined in your link shows using file pushed to the repo and decrypted with secret, it's different than what we are doing here, which is saving the secret to a file. Could you explain you concern more on this one?

suho: @blyscuit 
> This PR might be best to wait until the error is fixed.

Yes, we could wait. However, after some research, the issue is more related to `fastlane.deliver`. We could use `fastlane.pilot` instead (a temporary fix)

> The practices outlined in your link shows using file pushed to the repo and decrypted with secret, it's different than what we are doing here, which is saving the secret to a file.

As I mentioned, it is a practice that we could follow. For example, with [base 64 aprroach](https://docs.github.com/en/actions/security-guides/encrypted-secrets#storing-base64-binary-blobs-as-secrets) the approach is:
1. Encode the file or the key with base64 method
2. Add the base 64 encoded string to secrets
3. In the CI workflow, decode the secret and `echo` it to a file (appstorkey.p8)

Technically, it's the same but with GitHub's practice, we will do extra works: encode and decode the secret with base64 method

suho (approved):
LTGM

PR #353, opened 2022-09-23
[#333] As a developer, I can build the project with the adhoc and appstore methods with Fastlane Swift

### LINKED ISSUE #333: As a developer, I can build the project with the adhoc and appstore methods with Fastlane Swift
## Why

Currently, iOS developers need to learn at least basic Ruby to understand or write/update Fastlane configurations. To reduce the effort, we will create Fastlane in Swift language. So everyone in the iOS Chapter can use easier.

## Acceptance Criteria

- Use [Fastlane's Gym](https://docs.fastlane.tools/actions/gym/) action to build a project with the ad-hoc and appstore methods

vnntsu:
Resolves #333 

## What happened

- Provide 3 lanes for building:
  - AdHoc Staging
  - AdHoc Production
  - App Store

- Create `Build` to support preparing parameters before making the build.

## Insight

- Fastlane set `xcodebuildFormatter` with default option named `xcbeautify`, however, it doesn't work  after appending to  `xcodebuild` command. So I change it to `xcpretty`.
- For `exportOptions`, if we want to test on ios-template's generated app, we should comment current one, and uncomment `Constant.bundleId: "match \(type.method) \(Constant.bundleId)"` since our `bundle id` is not match with the design. Something like: Staging's bundle id should be `xxx.staging` but our ios-template's app is using without suffix `staging`.

## Proof Of Work

Screen Shot 2022-09-24 at 00 06 24

### DIFF
--- a/fastlane/Helpers/Build.swift
+++ b/fastlane/Helpers/Build.swift
@@ -1,0 +1,82 @@
+//
+//  Build.swift
+//  FastlaneRunner
+//
+//  Created by Su T. on 23/09/2022.
+//  Copyright © 2022 Nimble. All rights reserved.
+//
+
+final class Build {
+
+    static func adHoc(env: Environment) {
+        build(env: env, type: .adHoc)
+    }
+
+    static func appStore() {
+        build(env: .production, type: .appStore)
+    }
+
+    static private func build(
+        env: Environment,
+        type: BuildType
+    ) {
+        buildApp(
+            scheme: .userDefined(env.scheme),
+            clean: .userDefined(true),
+            outputName: .userDefined(env.productName),
+            includeSymbols: .userDefined(true),
+            includeBitcode: .userDefined(type == .appStore),
+            exportMethod: .userDefined(type.value),
+            exportOptions: .userDefined([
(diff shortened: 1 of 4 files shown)


fastlane/Helpers/Build.swift
```diff
@@ -0,0 +1,82 @@
...
+        buildApp(
+            scheme: .userDefined(env.scheme),
+            clean: .userDefined(true),
+            outputName: .userDefined(env.productName),
+            includeSymbols: .userDefined(true),
+            includeBitcode: .userDefined(type == .appStore),
+            exportMethod: .userDefined(type.value),
+            exportOptions: .userDefined([
+                // NOTE: bundleId should be `env.bundleId` instead of `Constant.bundleId`
+                // To test ios-template, uncomment right below
```
suho, line 32: If we test ios-template with `production` env, we don't need to uncomment this, right?

vnntsu: Yes, you're right. Since our staging bundle id for Staging also is the same as production, so only Staging needs that hard coded.

ducbm051291 (approved):

nkhanh44 (approved):

PR #359, opened 2022-09-27
[#336] As a developer, I can upload the build to App Store Connect with Fastlane Swift

vnntsu:
#336

## What happened

- Create `uploadToAppStore` function at `Distribution`
- Support making App Store API Key:
  - Add `appstoreConnectAPIKey` in `Secret`
  - Add `appStoreKeyId` and `appStoreKeyId` to `Constant`

## Insight

- Rearrange `Constant` `MARK` by alphabet order
- Create lane `buildAndUploadToAppStoreLane`
 
## Proof Of Work

Screen Shot 2022-09-27 at 16 51 08

Screen Shot 2022-09-27 at 16 58 50

### DIFF
--- a/fastlane/Constants/Constant.swift
+++ b/fastlane/Constants/Constant.swift
@@ -10,33 +10,33 @@ enum Constant {
 
     // MARK: - App Store
 
-    static let appStoreKeyId: String = "<#appStoreKeyId#>"
-    static let appStoreIssuerId: String = "<#appStoreIssuerId#>"
+    static let appStoreKeyId = "<#appStoreKeyId#>"
+    static let appStoreIssuerId = "<#appStoreIssuerId#>"
 
     // MARK: - Firebase
 
-    static let stagingFirebaseAppId: String = "<#stagingFirebaseAppId#>"
-    static let productionFirebaseAppId: String = "<#productionFirebaseAppId#>"
-    static let firebaseTesterGroups: String = "<#group1#>, <#group2#>"
+    static let stagingFirebaseAppId = "<#stagingFirebaseAppId#>"
+    static let productionFirebaseAppId = "<#productionFirebaseAppId#>"
+    static let firebaseTesterGroups = "<#group1#>, <#group2#>"
 
     // MARK: - Match
 
-    static let userName: String = "<#userName#>"
-    static let teamId: String = "<#teamId#>"
-    static let keychainName: String = "<#keychainName#>"
-    static let matchURL: String = "<#matchURL#>"
+    static let userName = "<#userName#>"
+    static let teamId = "<#teamId#>"
+    static let keychainName = "<#keychainName#>"
+    static let matchURL = "<#matchURL#>"
 
     // MARK: - Path
 
-    static let outputPath: String = "./Output"
-    static let buildPath: String = "\(outputPath)/Build"
-    static let derivedDataPath: String = "\(outputPath)/DerivedData"
+    static let outputPath = "./Output"
+    static let buildPath = "\(outputPath)/Build"
+    static let derivedDataPath = "\(outputPath)/DerivedData"
 
     // MARK: - Project
 
-    static let stagingBundleId: String = "{BUNDLE_ID_STAGING}"
-    static let productionBundleId: String = "{BUNDLE_ID_PRODUCTION}"
-    static let projectName: String = "{PROJECT_NAME}"
+    static let stagingBundleId = "{BUNDLE_ID_STAGING}"
+    static let productionBundleId = "{BUNDLE_ID_PRODUCTION}"
+    static let projectName = "{PROJECT_NAME}"
 
     // MARK: - Symbol
 
--- a/fastlane/Constants/Constant.swift
+++ b/fastlane/Constants/Constant.swift
@@ -8,6 +8,17 @@
 
 enum Constant {
 
+    // MARK: - App Store
+
+    static let appStoreKeyId: String = "<#appStoreKeyId#>"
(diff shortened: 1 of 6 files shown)


fastlane/Helpers/Distribution.swift
```diff
@@ -25,4 +25,27 @@ enum Distribution {
...
     }
+
+    static func uploadToAppStore() {
+        let environment: Constant.Environment = .production
+        let ipaPath = makeIpaPath(environment: environment)
+        appStoreConnectApiKey(
+            keyId: Constant.appStoreKeyId,
+            issuerId: Constant.appStoreIssuerId,
+            keyContent: .userDefined(Secret.appstoreConnectAPIKey),
+            isKeyContentBase64: .userDefined(true) // Check if the AppStore Connect API Key is base64
```
blyscuit, line 36:
```suggestion
            isKeyContentBase64: .userDefined(true) // Use AppStore Connect API Key in base64 encoded
```

blyscuit (approved):

fastlane/Helpers/Distribution.swift
vnntsu: @blyscuit We have two ways to procedure API key content, so I think it's better if the user checks to ensure they're using base64 or not. We don't need to force using base64 format, IMO.

fastlane/Helpers/Distribution.swift
```diff
@@ -25,4 +25,27 @@ enum Distribution {
             debug: .userDefined(true)
         )
     }
+
+    static func uploadToAppStore() {
+        let environment: Constant.Environment = .production
+        let ipaPath = makeIpaPath(environment: environment)
+        appStoreConnectApiKey(
```
suho, line 32: Let's separate this into a new function. I might use this for downloading the dSYM file. How about create

```swift
enum AppStoreAuthentication { static func connectAPIKey() }
```

fastlane/Helpers/Distribution.swift
```diff
@@ -25,4 +25,27 @@ enum Distribution {
...
+            appIdentifier: .userDefined(environment.bundleId),
+            ipa: .userDefined(ipaPath),
+            skipScreenshots: .userDefined(true),
+            skipMetadata: .userDefined(true),
+            force: .userDefined(true),
+            runPrecheckBeforeSubmit: .userDefined(false)
+        )
+    }
+
+    static private func makeIpaPath(environment: Constant.Environment) -> String {
```
suho, line 48: I would prefer

```swift
private static func makeIPAPath(...
```

fastlane/Helpers/Distribution.swift
```diff
@@ -25,4 +25,27 @@ enum Distribution {
             debug: .userDefined(true)
         )
     }
+
+    static func uploadToAppStore() {
+        let environment: Constant.Environment = .production
```
suho, line 30: How about this?
```swift
static func uploadToAppStore(environment: Constant.Environment = .production)
```

We can support upload staging build to AppStore as well, some project did this.

ducbm051291 (approved):
I think Su.V's suggestion is reasonable. The rest LGTM

fastlane/Constants/Constant.swift
```diff
@@ -8,30 +8,37 @@
...
+    // MARK: - App Store
+
+    static let appStoreKeyId: String = "<#appStoreKeyId#>"
+    static let appStoreIssuerId: String = "<#appStoreIssuerId#>"
+
+    // MARK: - Firebase
+
+    static let stagingFirebaseAppId: String = "<#stagingFirebaseAppId#>"
+    static let productionFirebaseAppId: String = "<#productionFirebaseAppId#>"
+    static let firebaseTesterGroups: String = "<#group1#>, <#group2#>"
```
minhnimble, line 20: Just curious, why don't we turn this into an array of Strings instead? 🤔
```suggestion
    static let firebaseTesterGroups: [String] = ["<#group1#>", "<#group2#>"]
```

ducbm051291: I think it's because in the final, it must convert to a string for Fastlane to send

vnntsu: @minhnimble Good concern, you can check the [doc](https://firebase.google.com/docs/app-distribution/ios/distribute-fastlane) of firebase distribution. It is a string, so for me, I don't want extra logic to join them when passing to the function. What do you think?

Screen Shot 2022-09-28 at 16 08 34

fastlane/Constants/Secret.swift
```diff
@@ -12,4 +12,6 @@ enum Secret {
     static let keychainPassword: String = "<#keychainPassword#>"
 
     static let firebaseCLIToken: String = "<#firebaseCLIToken#>"
+
+    static let appstoreConnectAPIKey = "<#appstoreConnectAPIKey#>"
```
blyscuit, line 16: This is the only constant without `: String`. I think we don't need the `: String` at all for these cases?

vnntsu: @blyscuit Updated as this [thread](https://nimble-co.slack.com/archives/C418DBL2W/p1664364433445829) too in [f419874](https://github.com/nimblehq/ios-templates/pull/359/commits/f419874843448853c24199fb176948596b703a8b) as

suho (approved):
Approved

fastlane/Constants/Constant.swift
minhnimble: noted, thanks for the sharing @vnntsu 👍

Issue #343, opened 2022-09-16
As a developer, I can register new device with Fastlane Swift


suho:
## Why

Currently, iOS developers need to learn at least basic Ruby to understand or write/update Fastlane configurations. To reduce the effort, we will create Fastlane in Swift language. So everyone in the iOS Chapter can use easier.

## Acceptance Criteria

- Use Fastlane's [register_devices](http://docs.fastlane.tools/actions/register_devices/#register_devices) action to register new device.

PR #365, opened 2022-09-30
[#343] As a developer, I can register new device with Fastlane Swift

nkhanh44:
#343 

## What happened

- Create a lane to register new device using `registerDevice()`.
 
## Insight

- Implement `registerDevice` lane that matches [current ruby implementation](https://github.com/nimblehq/ios-templates/blob/5664de6fc4a9da259295c4c1a1d3841d69b1dd72/fastlane/Fastfile#L89).
- `registerDevice` lane requires `device name`, `UDID`, `teamID` to register a new device.

### DIFF
--- a/fastlane/Fastfile.swift
+++ b/fastlane/Fastfile.swift
@@ -112,4 +112,20 @@ class Fastfile: LaneFile {
             devices: Constant.devices
         )
     }
+
+    // MARK: - Register device
+
+    func registerNewDeviceLane() {
+        let deviceName = prompt(text: "Enter the device name")
+        let deviceUDID = prompt(text: "Enter the device UDID:")
+
+        registerDevice(
+            name: deviceName,
+            udid: deviceUDID,
+            teamId: .userDefined(Constant.teamId)
+        )
+
+        Match.syncCodeSigning(type: .development, appIdentifier: [], isForce: true)
+        Match.syncCodeSigning(type: .adHoc, appIdentifier: [], isForce: true)
+    }
 }
--- a/fastlane/Fastfile.swift
+++ b/fastlane/Fastfile.swift
@@ -116,7 +116,7 @@ class Fastfile: LaneFile {
     // MARK: - Register device
 
     func registerNewDeviceLane() {
-        let deviceName = prompt(text: "Enter the device name")
+        let deviceName = prompt(text: "Enter the device name:")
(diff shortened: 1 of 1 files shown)


phongvhd93 (approved):

fastlane/Fastfile.swift
```diff
@@ -68,4 +68,20 @@ class Fastfile: LaneFile {
         Symbol.downloadFromAppStore(versionNumber: versionNumber, buildNumber: buildNumber)
         Symbol.uploadAppStoreToCrashlytics(versionNumber: versionNumber, buildNumber: buildNumber)
     }
+    
+    // MARK: - Register device
+
+    func registerNewDeviceLane() {
+        let deviceName = prompt(text: "Enter the device name")
```
Shayokh144, line 75: To make it similar with the next line
```suggestion
        let deviceName = prompt(text: "Enter the device name:")
```

vnntsu (approved):

fastlane/Fastfile.swift
```diff
@@ -68,4 +68,20 @@ class Fastfile: LaneFile {
...
+        let deviceUDID = prompt(text: "Enter the device UDID:")
+
+        registerDevice(
+            name: deviceName,
+            udid: deviceUDID,
+            teamId: .userDefined(Constant.teamId)
+        )
+
+        Match.syncCodeSigning(type: .development, appIdentifier: [], isForce: true)
+        Match.syncCodeSigning(type: .adHoc, appIdentifier: [], isForce: true)
```
Shayokh144, line 85: can we provide user the option to decide if they want automatic renew of provisioning profiles or not?

fastlane/Fastfile.swift
nkhanh44: Good catch! I updated at https://github.com/nimblehq/ios-templates/pull/365/commits/9b582f73aa4a0ceb1c83f52fc7d69fcb26a3b1c1.

fastlane/Fastfile.swift
suho: @Shayokh144 Because we are registering new devices, so using `isForce: true` here to regenerate provisioning profiles. Normally, we don't need to use `isForce` :D.

But I will create other lanes for developers to run sync code signing.

Shayokh144 (approved):

Issue #324, opened 2022-08-19
Use Moya as the main Network Layer 


blyscuit:
### Benefit
- We can stop maintaining our network layer.
- Every projects have the same network layer.
- Knowledge transfer between projects.
- Ease of onboarding because Moya is easier to read than Alamofire.

## Acceptance Criteria

- Add Moya to the project.
- Remove redundant network layer code.

## Resources

https://github.com/Moya/Moya/tree/master/docs

blyscuit:
Should we adopt Moya?

/polls Yes, let's go 🚀 'No, I love current network layer'

blyscuit:
Please vote in the [discussion](https://github.com/nimblehq/ios-templates/discussions/325)

ducbm051291:
Current template direction is explicitly Alamofire-based in both code and docs

PR #394, opened 2022-11-28
[#324] Use Moya as the main Network Layer

Shayokh144:
https://github.com/nimblehq/ios-templates/issues/324

## What happened

Replace `Alamofire` with `Moya` as main network layer of the template.
 
## Insight

- Remove `Alamofire` from Pod
- Add `Moya` in Pod
- Update networking code to use `Moya`
- Add comments for better readability
- Test this branch by creating a separate app from it
- Remove common `RequestConfiguration` as it is not needed, developers can create their own request configuration `enum` extending Moya's `TargetType` 
 
## Proof Of Work
- Created a demo app from this branch, request-response working fine
- Create different configurations independently 

- **Sample configuration 1:**
Screenshot 2022-12-01 at 5 40 35 PM

- **Sample configuration 2:**
Screenshot 2022-12-01 at 5 39 23 PM

- **Call using any configuration with any decodable type**
Screenshot 2022-12-01 at 5 45 46 PM

- **Write test with stubbed provider**

Screenshot 2022-12-02 at 4 21 01 PM


{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
```diff
@@ -12,33 +11,12 @@ protocol NetworkAPIProtocol {
...
-                } catch {
-                    observer.on(.error(error))
-                }
-                observer.on(.completed)
-                return Disposables.create()
-            }
-        }
-        .asSingle()
+        provider.rx.request(configuration)
+            .filterSuccessfulStatusAndRedirectCodes()
```
ducbm051291, line 19: Correct me if I'm wrong. I don't think that we need to filter the status code here. If there's a server error with code 5xx, will this request have no response? 

`filterSuccessfulStatusAndRedirectCodes() filters status codes that are in the 200-300 range.`

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
```diff
@@ -2,38 +2,34 @@
...
-    var baseURL: String { get }
-
-    var endpoint: String { get }
-
-    var method: HTTPMethod { get }
-
-    var url: URLConvertible { get }
-
-    var parameters: Parameters? { get }
+    /// Add cases for each endpoint
```
blyscuit, line 10: No need for documentation comments, it's ok to see on Moya's repository.

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
```diff
@@ -2,38 +2,34 @@
...
+    /// Build and Return HTTP task for each endpoint case
+    var task: Moya.Task { .requestPlain }
 
-    var parameters: Parameters? { nil }
+    /// Return the appropriate HTTP headers for every endpoint case
+    var headers: [String: String]? { ["Content-Type": "application/json"] }
 
-    var headers: HTTPHeaders? { nil }
+    /// Return stub/mock data for use in testing. Default is `Data()`
+    var sampleData: Data { Data() }
```
blyscuit, line 31: No need to include optional vars.

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
```diff
@@ -2,38 +2,34 @@
...
 
-    var interceptor: RequestInterceptor? { nil }
+    /// Return the type of validation to perform on the request. Default is `.none`.
+    var validationType: ValidationType { .successCodes }
```
blyscuit, line 34: This has a default value, user of the repo can modify this by themself.

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
```diff
@@ -2,38 +2,34 @@
...
-
-    var parameters: Parameters? { get }
+    /// Add cases for each endpoint
+}
 
-    var encoding: ParameterEncoding { get }
+extension RequestConfiguration: TargetType {
 
-    var headers: HTTPHeaders? { get }
+    /// Return base URL for a target
```
blyscuit, line 15: No need for documentation comments, it's ok to see on Moya's repository. And for other comments here.

blyscuit (commented):
Please check David's comment as well.

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
```diff
@@ -2,38 +2,34 @@
 //  RequestConfiguration.swift
 //
 
-import Alamofire
 import Foundation
+import Moya
 
-protocol RequestConfiguration {
+enum RequestConfiguration {
```
vnntsu, line 8: `RequestConfiguration` was a protocol to define a blueprint for every request. It couldn't be an enum or struct like this since we will have a specific config for each group of APIs based on the endpoint, e.g `user` -> `UserRequestConfiguration`.

It's also great if we keep `RequestConfiguration` as its responsibility with `Alamofire`, which means it's like a wrapper of Moya. Let's say we will not use Moya and change to another one in the future, so we don't need to update every "RequestConfiguration" to conform to the new lib. Conversely, we're going to import Moya to every request config to be able conform `TargetType` (and yeah, now it should be named `ABCTarget`).

Shayokh144: @vnntsu I am a bit confused about this.. Do you suggest to keep the `RequestConfiguration` protocol unchanged?

ducbm051291: @vnntsu We're using Moya, so I think using Enum will be fine. In case we have different groups of APIs, we can create another RequestConfiguration like UserRequestConfiguration, as long as we implement the protocol TargetType

vnntsu: @Shayokh144 I suggest keeping `RequestConfiguration` protocol as a wrapper of TargetType; otherwise, we remove it. 
The enum `RequestConfiguration` is redundancy. We will have specific request configs such as `UserRequestConfiguration` or `OtherRequestConfiguration`, etc.

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
```diff
@@ -12,33 +11,11 @@ protocol NetworkAPIProtocol {
 }
 
 extension NetworkAPIProtocol {
-
     func request<T: Decodable>(
-        session: Session,
-        configuration: RequestConfiguration,
-        decoder: JSONDecoder
+        provider: MoyaProvider<RequestConfiguration>,
```
vnntsu, line 15: As I mentioned above, if I have `UserRequestConfiguation`, how can I make a request? This is not a core because it's only for the enum `RequestConfiguration`

vnntsu (changes requested):

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
suho: Before Moya, our `RequestConfiguration` is `TargetType`, and it's a wrapper for Alamofire as we are moving to `Moya`, so I prefer to remove `RequestConfiguration`.

@vnntsu did raise a good point about changing to another networking library in the feature. We might need to update every `RequestConfiguration` to conform to the new library. In this case, I prefer to have a typealias:

```swift
typealias TargetType = Moya.TargetType
```

Let's say if we decide to remove Moya and go back to Alamofire, then we just need to implement the protocol `TargetType` (with baseURL, method...) in the network layer.

P.s: The reason I suggest `typealias TargetType = Moya.TargetType` instead of `typealias RequestConfiguration = Moya.TargetType` because it's more clear, using `RequestConfiguration` might confuse developers when we are using Moya.

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
suho: This's a good catch!

@Shayokh144 You might need to think about how we create different `RequestConfiguration`  (or `TargetType`) and handle it in the core service. On the other hand, it's better to show about testing code with Moya in the POW section

ducbm051291: So need a wrapper for the TargetType

`protocol RequestConfiguration: TargetType {}`

Shayokh144: @vnntsu @suho I have added a new commit [here](https://github.com/nimblehq/ios-templates/pull/394/commits/d37a1e0af77dc98bfedf7be31ded6a3ed901aee4). Now the core network api's are generic and reusable.

{PROJECT_NAME}/Sources/Data/NetworkAPI/NetworkAPI.swift
```diff
@@ -2,23 +2,19 @@
...
 
-import Alamofire
 import Foundation
+import Moya
 import RxSwift
 
 final class NetworkAPI: NetworkAPIProtocol {
 
-    private let decoder: JSONDecoder
+    private let provider: MoyaProvider<MultiTarget>
```
blyscuit, line 11: Please work with @suho to clarify this [comment](https://github.com/nimblehq/ios-templates/pull/394#discussion_r1035608474)

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
```diff
@@ -2,43 +2,20 @@
...
+    func performRequest<T: Decodable>(_ configuration: TargetType, for type: T.Type) -> Single<T>
 }
 
 extension NetworkAPIProtocol {
-
     func request<T: Decodable>(
-        session: Session,
-        configuration: RequestConfiguration,
-        decoder: JSONDecoder
+        provider: MoyaProvider<MultiTarget>,
```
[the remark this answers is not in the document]
Shayokh144: @blyscuit Second one is working:


Screenshot 2022-12-02 at 1 48 12 PM

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/NetworkAPIProtocol.swift
```diff
@@ -2,43 +2,20 @@
...
-                } catch {
-                    observer.on(.error(error))
-                }
-                observer.on(.completed)
-                return Disposables.create()
-            }
-        }
-        .asSingle()
+        provider.rx.request(MultiTarget(configuration))
+            .map(T.self)
```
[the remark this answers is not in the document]
Shayokh144: @blyscuit decoder added in this [commit](https://github.com/nimblehq/ios-templates/pull/394/commits/f7d5b3c7c3bb747646829b3f884e5338170260d7)

{PROJECT_NAME}/Sources/Data/NetworkAPI/Core/RequestConfiguration.swift
Shayokh144: @suho typealias added and RequestConfiguration is deleted

blyscuit (approved):
🚀

ducbm051291 (approved):

{PROJECT_NAME}/Sources/Data/NetworkAPI/NetworkAPI.swift
suho: @Shayokh144 Regarding [this discussion](https://nimble-co.slack.com/archives/C418DBL2W/p1670300978623669), it's better to remove `NetworkAPI` and `NetworkAPIProtocol` in this PR

vnntsu:
@Shayokh144 we will have this initiative [Replace-Network-Layer-with-Moya](https://www.notion.so/nimblehq/Replace-Network-Layer-with-Moya-e303c73760f04499bbee2357330601a2) to work on before making the change to our ios templates. So I'll close this PR.
cc: @nimblehq/ios-chapter

Issue #399, opened 2022-12-02
Fastlane Swift - Fix Cannot load such file -- ./fastlane/Constants/Constants


nkhanh44:
## Issue

- Danger cannot load the file `./fastlane/Constants/Constants` error.

![Screen Shot 2022-11-29 at 23 10 33](https://user-images.githubusercontent.com/25881847/205216866-f2cce984-a917-407c-a3b8-590e9b29fcfc.png)
 
## Expected

In Dangerfile:
![Screenshot 2022-12-02 at 11 53 29](https://user-images.githubusercontent.com/25881847/205217654-307abd3c-1a29-47a7-9fbe-3b89fbdbafbd.png)

- Set it to the correct path `./fastlane/Constants/Constant`.
 
## Steps to reproduce

1. From the migrate-develop branch
2. Push a PR and observe the error result.

ducbm051291:
Danger requires Constants.rb and DeliverableConstants.rb to run but in the middle of migrating Fastlane Swift, we replaced them with Constant.swift. So let's restore those two files and finish migrating Fastlane Swift. 

While investigating this case, I found that we're missing replace two lanes in the pull request review workflow.

![image](https://user-images.githubusercontent.com/24598204/208075109-7d270287-4c78-439c-8a50-09c69cd8e261.png)

![image](https://user-images.githubusercontent.com/24598204/208075124-320585b4-207a-40c0-ab6d-60d60e3d2a58.png)

suho:
@ducbm051291 For this issue, I think you can manually define all constants into the Dangerfile, you don't need to import (require) the Constants file 😉

ducbm051291:
> @ducbm051291 For this issue, I think you can manually define all constants into the Dangerfile, you don't need to import (require) the Constants file 😉

Ok let me give a try 🙇‍♂️

PR #400, opened 2022-12-09
[#399] Fastlane Swift - Fix Cannot load such file ./fastlane/Constants/Constants

ducbm051291:
- #399 

## What happened

- Danger cannot load the file ./fastlane/Constants/Constants due to an incorrect path. 
 
## Insight

- Update Dangerfile with the correct path ./fastlane/Constants/Constant.
 
## Proof Of Work

Screenshot 2022-12-09 at 09 15 39

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require './fastlane/Constants/Constant'
+require './Constants'
 
 # Warn when there is a big PR
 warn("This pull request is quite big (#{git.lines_of_code} lines changed), please consider splitting it into multiple pull requests.") if git.lines_of_code > 500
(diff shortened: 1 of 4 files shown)


blyscuit (approved):

nkhanh44 (approved):

Dangerfile
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require './fastlane/Constants/Constants'
+require './fastlane/Constants/Constant'
```
suho, line 3: Does this work? It's a Swift file, while this Dangerfile contains Ruby code.

Please try it by generating a project and create a pull request and update the POW

ducbm051291: You're correct, this issue is not simple as I expected, I changed approach and will create another PR

PR #413, opened 2022-12-20
[#399] [Bug] Fastlane Swift - Fix Cannot load such file -- ./fastlane/Constants/Constants

ducbm051291:
- #399 

## What happened

- Danger cannot load the file ./fastlane/Constants/Constants due to an incorrect path. 
 
## Insight

- Update defining constants in Danger file
- Update Automatic pull request review workflow using the new swift lane.
- Update Automatic pull request review workflow using macOS-12 instead of the latest. Ref: https://github.com/actions/runner-images/issues/6384

Note: We may update DangerSwift in the future for consistent with FastlaneSwift

## Proof Of Work

Screenshot 2022-12-20 at 20 59 18


fastlane/Constants/Constants.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+class Constants
```
suho, line 3: Please don't create this file, you need to define the constants directly in the Dangerfile

suho (changes requested):

fastlane/Constants/Constants.rb
ducbm051291: But how do we know the application's name before the user run make.sh command?

ducbm051291: Updated in ba738a1

Dangerfile
```diff
@@ -23,7 +19,10 @@ swiftlint.lint_files(
   additional_swiftlint_args: '--strict'
 )
 
-xcresultPath = "#{Constants.TEST_OUTPUT_DIRECTORY_PATH}/#{Constants.TESTS_SCHEME}.xcresult"
+xcresultPath = "./fastlane/test_output/{PROJECT_NAME} Staging.xcresult"
+scheme = "{PROJECT_NAME} Staging"
+workspace = "./{PROJECT_NAME}.xcworkspace"
+outputDirectory = "./fastlane/xcov_output"
```
suho, line 25: How about better naming for this?
```suggestion
xcovOutputDirectory = "./fastlane/xcov_output"
```

make.sh
```diff
@@ -189,12 +189,6 @@ rm -rf .github/workflows/test_install_script.yml
 rm -f .git/index
 git reset
 
-if [[ -z "${CI}" ]]; then
```
suho, line 192: Why you removed this?

ducbm051291: Cause we don't use deliverable_setup.sh anymore, so I removed this. I will restore the rm part

suho: I think we still need to use it, but with Swift Fastlane, please revert the changes

ducbm051291: Done

suho: @ducbm051291 Please revert all removed changes:
- make.sh
- set_up_test_testflight.sh
- deliverable_setup.sh

If the workflow does not work, please create a new task for that

ducbm051291: Got it ;)

suho (approved):

vnntsu (approved):

Issue #330, opened 2022-08-23
Modify signing automatically when generating a project


blyscuit:
## Why

When generating a new project from make.sh, the `Signing` is `automatic signing`. This does not change when running `fastlane match`.

This causes issues:
- CICD fail if the user did not modify Signing manually.
- CICD fail if the user modify Signing manually but push incorrect settings.
- Local build fail the user did not modify Signing manually. 

Example test script for local run:
<img width="1512" alt="Screen Shot 2565-08-23 at 13 01 04" src="https://user-images.githubusercontent.com/6356137/186083742-3d7fe6c8-a0be-4dfc-b2e2-e773f4e21ed5.png">
<img width="1136" alt="Screen Shot 2565-08-23 at 13 00 51" src="https://user-images.githubusercontent.com/6356137/186083764-077a9586-be49-4964-9a9a-966db90ec391.png">

```
cd ..
rm -rf TemplateApp
cp -R ./ios-templates/ ./TemplateApp
cd TemplateApp
sh make.sh --bundle-id co.nimblehq.ios.templates --bundle-id-staging co.nimblehq.ios.templates.staging --project-name TemplateApp

find ./ -type f -exec sed -i -e 's/git@github.com:{organization}\/{repo}.git/git@github.com:nimblehq\/match-certificates.git/g' {} \;

bundle install
bundle exec fastlane sync_appstore_signing
bundle exec fastlane build_and_upload_appstore_app
```
## Acceptance Criteria  

- Add new lane using this example
```
  update_code_signing_settings(
     use_automatic_signing: false,
     team_id: ENV["sigh_#{Constants.BUNDLE_ID_PRODUCTION}_appstore_team-id"],
     profile_name: ENV["sigh_#{Constants.BUNDLE_ID_PRODUCTION}_appstore_profile-name"],
     code_sign_identity: 'iPhone Distribution'
   )
```
- New lane should behave correctly for:
- Production App Store
- Production Adhoc
- Staging Adhoc

- Use new lane for CICD when `match`.
- Add document for new lane in the Firebase section https://github.com/nimblehq/ios-templates/blob/c06525db7c05e4c70e6682648a9fd8aa741b0ed9/.github/wiki/Fastlane.md.

## Who Benefits?

Users.

PR #450, opened 2023-03-01
[#330] Modify signing automatically when generating a project

edgarss:
- Close #https://github.com/nimblehq/ios-templates/issues/330

## What happened 👀

When generating a new project from make.sh, the Signing is automatic signing. This does not change when running fastlane match.

This causes issues:

CICD fail if the user did not modify Signing manually.
CICD fail if the user modify Signing manually but push incorrect settings.
Local build fail the user did not modify Signing manually.

## Insights

Added `updateCodeSigningSettings` to update the xcode project with synced match provisioning profiles after each match sync

Added syncDevelopmentProductionCodeSigningLane since there was no lane to generate a provisioning profile for Debug Production configuration.

## Proof Of Work 📹

Screenshot 2566-03-01 at 11 45 06

Screenshot 2566-03-01 at 11 23 47

### DIFF
--- a/.github/workflows/test_upload_build_to_test_flight.yml
+++ b/.github/workflows/test_upload_build_to_test_flight.yml
@@ -56,7 +56,7 @@ jobs:
       run: bundle exec fastlane setUpTestProject
 
     - name: Update Provision Profile
-      run: bundle exec fastlane updateProvisionSettings
+      run: bundle exec fastlane syncAppStoreCodeSigning
       env:
         MATCH_PASSWORD: ${{ secrets.MATCH_PASS }}
 
--- a/.github/workflows/test_upload_build_to_test_flight.yml
+++ b/.github/workflows/test_upload_build_to_test_flight.yml
@@ -7,6 +7,7 @@ name: Test Upload Build to TestFlight
 ### API_KEY_ID
 ### ISSUER_ID
 ### APPSTORE_CONNECT_API_KEY
+### TEAM_ID
 
 on:
   pull_request
@@ -51,6 +52,7 @@ jobs:
         MATCH_REPO: ${{ secrets.MATCH_REPO }}
         API_KEY_ID: ${{ secrets.API_KEY_ID }}
         ISSUER_ID: ${{ secrets.ISSUER_ID }}
+        TEAM_ID: ${{ secrets.TEAM_ID }}
 
     - name: Set Up Test Project for App Store
       run: bundle exec fastlane setUpTestProject
--- a/fastlane/Fastfile.swift
+++ b/fastlane/Fastfile.swift
@@ -12,11 +12,19 @@ class Fastfile: LaneFile {
 
     // MARK: - Code signing
 
-    func syncDevelopmentCodeSigningLane() {
+    func syncDevelopmentStagingCodeSigningLane() {
         desc("Sync the Development match signing for the Staging build")
         Match.syncCodeSigning(
             type: .development,
             appIdentifier: [Constant.stagingBundleId]
+        )
+    }
+    
+    func syncDevelopmentProductionCodeSigningLane() {
+        desc("Sync the Development match signing for the Staging build")
+        Match.syncCodeSigning(
+            type: .development,
+            appIdentifier: [Constant.productionBundleId]
         )
     }
 
--- a/fastlane/Fastfile.swift
+++ b/fastlane/Fastfile.swift
@@ -21,7 +21,7 @@ class Fastfile: LaneFile {
     }
     
     func syncDevelopmentProductionCodeSigningLane() {
-        desc("Sync the Development match signing for the Staging build")
+        desc("Sync the Development match signing for the Production build")
         Match.syncCodeSigning(
             type: .development,
             environment: .production
@@ -88,7 +88,7 @@ class Fastfile: LaneFile {
     }
 
     func buildProductionAndUploadToFirebaseLane() {
-        desc("Build Staging app and upload to Firebase")
+        desc("Build Production app and upload to Firebase")
 
         setAppVersion()
         bumpBuild()
--- a/fastlane/Fastfile.swift
+++ b/fastlane/Fastfile.swift
@@ -145,20 +145,6 @@ class Fastfile: LaneFile {
         )
     }
 
-    func updateProvisionSettingsLane() {
-        desc("Update Provision Profile")
-        syncAppStoreCodeSigningLane()
-        updateCodeSigningSettings(
-            path: Constant.projectPath,
-            useAutomaticSigning: .userDefined(false),
-            teamId: .userDefined(EnvironmentParser.string(key: "sigh_\(Constant.productionBundleId)_appstore_team-id")),
-            codeSignIdentity: .userDefined("iPhone Distribution"),
-            profileName: .userDefined(EnvironmentParser.string(
-                key: "sigh_\(Constant.productionBundleId)_appstore_profile-name"
-            ))
-        )
-    }
-
     func setUpTestProjectLane() {
         desc("Disable Exempt Encryption")
         Test.disableExemptEncryption()
--- a/fastlane/Fastfile.swift
+++ b/fastlane/Fastfile.swift
@@ -16,7 +16,7 @@ class Fastfile: LaneFile {
         desc("Sync the Development match signing for the Staging build")
         Match.syncCodeSigning(
             type: .development,
-            appIdentifier: [Constant.stagingBundleId]
+            environment: .staging
         )
     }
     
@@ -24,7 +24,7 @@ class Fastfile: LaneFile {
         desc("Sync the Development match signing for the Staging build")
         Match.syncCodeSigning(
             type: .development,
-            appIdentifier: [Constant.productionBundleId]
+            environment: .production
         )
     }
 
@@ -32,7 +32,7 @@ class Fastfile: LaneFile {
         desc("Sync the Ad Hoc match signing for the Staging build")
         Match.syncCodeSigning(
             type: .adHoc,
-            appIdentifier: [Constant.stagingBundleId]
+            environment: .staging
         )
     }
 
@@ -40,7 +40,7 @@ class Fastfile: LaneFile {
         desc("Sync the Ad Hoc match signing for the Production build")
         Match.syncCodeSigning(
             type: .adHoc,
-            appIdentifier: [Constant.productionBundleId]
+            environment: .production
         )
     }
 
@@ -48,7 +48,7 @@ class Fastfile: LaneFile {
         desc("Sync the App Store match signing for the Production build")
         Match.syncCodeSigning(
             type: .appStore,
-            appIdentifier: [Constant.productionBundleId]
+            environment: .production
         )
     }
 
@@ -170,8 +170,8 @@ class Fastfile: LaneFile {
             teamId: .userDefined(Constant.teamId)
         )
 
-        Match.syncCodeSigning(type: .development, appIdentifier: [], isForce: true)
-        Match.syncCodeSigning(type: .adHoc, appIdentifier: [], isForce: true)
+        Match.syncCodeSigning(type: .development, environment: .staging, isForce: true)
+        Match.syncCodeSigning(type: .adHoc, environment: .staging, isForce: true)
     }
 
     // MARK: - Utilities
--- a/fastlane/Helpers/Match.swift
+++ b/fastlane/Helpers/Match.swift
@@ -66,13 +66,13 @@ extension Match {
     }
     
     enum MatchType: String {
-        
+
         case development
         case adHoc = "adhoc"
         case appStore = "appstore"
-        
+
         var value: String { return rawValue }
-        
+
         var method: String {
             switch self {
             case .development: return "Development"
--- a/fastlane/Helpers/Match.swift
+++ b/fastlane/Helpers/Match.swift
@@ -12,7 +12,7 @@ enum Match {
         if isCi() {
             Keychain.create()
             match(
-                type: type.value,
+                type: type.match,
                 readonly: .userDefined(!isForce),
                 appIdentifier: [environment.bundleId],
                 username: .userDefined(Constant.userName),
@@ -24,7 +24,7 @@ enum Match {
             )
         } else {
             match(
-                type: type.value,
+                type: type.match,
                 readonly: .userDefined(!isForce),
                 appIdentifier: [environment.bundleId],
                 username: .userDefined(Constant.userName),
--- a/fastlane/Helpers/Match.swift
+++ b/fastlane/Helpers/Match.swift
@@ -32,8 +32,8 @@ enum Match {
                 gitUrl: Constant.matchURL,
                 force: .userDefined(isForce)
             )
-            updateCodeSigning(type: type, appIdentifier: appIdentifier)
         }
+        updateCodeSigning(type: type, appIdentifier: appIdentifier)
     }
     
     static func updateCodeSigning(type: MatchType, appIdentifier: [String]) {
--- a/fastlane/Helpers/Match.swift
+++ b/fastlane/Helpers/Match.swift
@@ -7,7 +7,7 @@
 //
 
 enum Match {
-    
+
     static func syncCodeSigning(type: MatchType, environment: Environment, isForce: Bool = false) {
         if isCi() {
             Keychain.create()
--- a/fastlane/Helpers/Match.swift
+++ b/fastlane/Helpers/Match.swift
@@ -40,34 +40,34 @@ enum Match {
         // Update Code signing from automatic to manual
         switch type {
         case .development:
-            updateCodeSigningSettings(
-                path: Constant.projectPath,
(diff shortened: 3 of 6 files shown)


fastlane/Helpers/Match.swift
```diff
@@ -33,6 +33,51 @@ enum Match {
...
+                codeSignIdentity: .userDefined("iPhone Developer"),
+                profileName: .userDefined("match AdHoc \(Constant.stagingBundleId)")
+            )
+        case .adHoc:
+            if appIdentifier.first == Constant.productionBundleId {
+                updateCodeSigningSettings(
+                    path: Constant.projectPath,
+                    useAutomaticSigning: .userDefined(false),
+                    targets: .userDefined([Constant.projectName]),
+                    buildConfigurations: .userDefined(["Debug Production"]),
```
ducbm051291, line 57: Correct me if I'm wrong, but I think this is supposed to be Release Production

fastlane/Helpers/Match.swift
```diff
@@ -33,6 +33,51 @@ enum Match {
...
+    
+    static func updateCodeSigning(type: MatchType, appIdentifier: [String]) {
+        // Update Code signing from automatic to manual
+        switch type {
+        case .development:
+            updateCodeSigningSettings(
+                path: Constant.projectPath,
+                useAutomaticSigning: .userDefined(false),
+                targets: .userDefined([Constant.projectName]),
+                buildConfigurations: .userDefined(["Release Staging"]),
```
ducbm051291, line 47: Should this be Debug Staging?

fastlane/Helpers/Match.swift
```diff
@@ -33,6 +33,51 @@ enum Match {
...
+                )
+            }
+        case .appStore:
+            updateCodeSigningSettings(
+                path: Constant.projectPath,
+                useAutomaticSigning: .userDefined(false),
+                targets: .userDefined([Constant.projectName]),
+                buildConfigurations: .userDefined(["Release Production"]),
+                codeSignIdentity: .userDefined("iPhone Distribution"),
+                profileName: .userDefined("match AppStore \(Constant.productionBundleId)")
```
vnntsu, line 78: I prefer to define `AdHoc` and `AppStore` as `MatchType`'s attributes such as `type`:
```swift
    enum MatchType: String {
        ...
        var type: String {
            switch self {
            case .development: return "Development"
            case .adHoc: return "AdHoc"
            case .appStore: return "AppStore"
            }
        }
    }
```

And a private function to make `profileName` so we don't duplicate many `match` words.

fastlane/Helpers/Match.swift
```diff
@@ -33,6 +33,51 @@ enum Match {
...
+            updateCodeSigningSettings(
+                path: Constant.projectPath,
+                useAutomaticSigning: .userDefined(false),
+                targets: .userDefined([Constant.projectName]),
+                buildConfigurations: .userDefined(["Release Staging"]),
+                codeSignIdentity: .userDefined("iPhone Developer"),
+                profileName: .userDefined("match AdHoc \(Constant.stagingBundleId)")
+            )
+        case .adHoc:
+            if appIdentifier.first == Constant.productionBundleId {
```
vnntsu, line 52: I think you're putting the wrong order for this if else. The `If` should be for production stuff otherwise it's ad-hoc.

fastlane/Helpers/Match.swift
```diff
@@ -33,6 +33,51 @@ enum Match {
...
+        // Update Code signing from automatic to manual
+        switch type {
+        case .development:
+            if appIdentifier.first == Constant.productionBundleId {
+                updateCodeSigningSettings(
+                    path: Constant.projectPath,
+                    useAutomaticSigning: .userDefined(false),
+                    targets: .userDefined([Constant.projectName]),
+                    buildConfigurations: .userDefined(["Debug Production"]),
+                    codeSignIdentity: .userDefined("iPhone Developer"),
```
blyscuit, line 49: I think we can declare `iPhone Developer` as the Type enum var
```
        var codeSignIdentity: String {
             switch self {
             case .development: return "iPhone Developer"
             case .adHoc, . appStore: return "iPhone Distribution"
             }
         } 
```

fastlane/Helpers/Match.swift
```diff
@@ -33,6 +33,51 @@ enum Match {
...
+    static func updateCodeSigning(type: MatchType, appIdentifier: [String]) {
+        // Update Code signing from automatic to manual
+        switch type {
+        case .development:
+            if appIdentifier.first == Constant.productionBundleId {
+                updateCodeSigningSettings(
+                    path: Constant.projectPath,
+                    useAutomaticSigning: .userDefined(false),
+                    targets: .userDefined([Constant.projectName]),
+                    buildConfigurations: .userDefined(["Debug Production"]),
```
blyscuit, line 48: I think we can define buildConfiguration as enum var
```
        var buildConfiguration: String {
             switch self {
             case .development: return "Debug"
             case .adHoc, .appStore: return "Release"
             }
         } 
``` 
and here we change 
```suggestion
                    buildConfigurations: .userDefined(["\(type.buildConfiguration) Production"]),
```

blyscuit (commented):
Thank you for the big help. This would reduce headache with code signing by a lot. 🙏 

Please check the TestFlight testing lane, something (signing related possibly) seems to be wrong with it 🤔

ducbm051291:
> Thank you for the big help. This would reduce headache with code signing by a lot. 🙏
> 
> Please check the TestFlight testing lane, something (signing related possibly) seems to be wrong with it 🤔

The TestFlight testing lane call `updateCodeSigning` lane but it was removed, @edgarss please update this with call `syncAppStoreCodeSigning` lane 😀

fastlane/Fastfile.swift
```diff
@@ -142,20 +150,6 @@ class Fastfile: LaneFile {
         )
     }
 
-    func updateProvisionSettingsLane() {
```
blyscuit, line 145: Since this is removed, we will need to modify the testing lane https://github.com/nimblehq/ios-templates/blob/bb6ab46d0f41a6048600cfc921d788850418b89a/.github/workflows/test_upload_build_to_test_flight.yml to include the new match and provision methods too.

Now we need to run `syncAppStoreCodeSigningLane` on the workflow instead of `updateProvisionSettingsLane` 🙏

fastlane/Helpers/Match.swift
```diff
@@ -26,17 +26,42 @@ enum Match {
...
+            buildConfigurations: .userDefined(["\(type.buildConfiguration) \(environment.rawValue)"]),
+            codeSignIdentity: .userDefined(type.codeSignIdentity),
+            profileName: .userDefined("match \(type.method) \(environment.bundleId)")
+        )
     }
 }
 
 extension Match {
+    
+    enum Environment: String {
```
blyscuit, line 54: I suggest returning array here in case some project has more than one bundle id for each environment.

edgarss: I still think its better to call `syncCodeSigning` for each environment seperately like we have been doing till now. Especially if we plan to support multiple apple accounts on the same project

vnntsu (commented):
@edgarss The code looks good! Please check failed CI.

ducbm051291:
@vnntsu I've checked, and we need to include the team Id in updateCodeSigning. I updated the hard-code teamId when running set_up_test_testflight.sh and CI worked perfectly

fastlane/Fastfile.swift
```diff
@@ -12,35 +12,43 @@ class Fastfile: LaneFile {
...
         desc("Sync the Development match signing for the Staging build")
         Match.syncCodeSigning(
             type: .development,
-            appIdentifier: [Constant.stagingBundleId]
+            environment: .staging
+        )
+    }
+    
+    func syncDevelopmentProductionCodeSigningLane() {
+        desc("Sync the Development match signing for the Staging build")
```
Shayokh144, line 24: here `production` environment is used, so please update the description:
desc("Sync the Development match signing for the Production build")

ducbm051291: Nice catch

fastlane/Helpers/Match.swift
```diff
@@ -26,24 +26,72 @@ enum Match {
...
+            switch self {
+            case .staging: return Constant.stagingBundleId
+            case .production: return Constant.productionBundleId
+            }
+        }
+    }
+    
     enum MatchType: String {
-
+
```
blyscuit, line 69: Please fix line break indent.
```suggestion

```

fastlane/Helpers/Match.swift
```diff
@@ -26,24 +26,72 @@ enum Match {
...
+    }
+    
     enum MatchType: String {
-
+        
         case development
         case adHoc = "adhoc"
         case appStore = "appstore"
-
+
```
blyscuit, line 73:
```suggestion

```

set_up_test_testflight.sh
```diff
@@ -1,12 +1,14 @@
...
 
+readonly CONSTANT_TEAM_ID="<#teamId#>"
 readonly CONSTANT_API_KEY_ID="<#API_KEY_ID#>"
 readonly CONSTANT_ISSUER_ID="<#ISSUER_ID#>"
 readonly CONSTANT_MATCH_REPO="git@github.com:{organization}\/{repo}.git"
 
 readonly WORKING_DIR=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
 MATCH_REPO_ESCAPED=$(echo "${MATCH_REPO//\//\\\/}")
 
+LC_ALL=C find $WORKING_DIR -type f -exec sed -i "" "s/$CONSTANT_TEAM_ID/4TWS7E2EPE/g" {} +
```
blyscuit, line 11: I think it's best to use value from GitHub Secret for this value 🙏

blyscuit (approved):
Few comments. Otherwise looks good.

ducbm051291 (approved):

fastlane/Helpers/Match.swift
```diff
@@ -7,14 +7,14 @@
 //
 
 enum Match {
-
-    static func syncCodeSigning(type: MatchType, appIdentifier: [String], isForce: Bool = false) {
+
```
blyscuit, line 10: Please remove this empty indent.

set_up_test_testflight.sh
```diff
@@ -1,12 +1,14 @@
 echo "import('./Tests/Fastfile')" | cat - fastlane/Fastfile | tee fastlane/Fastfile &> /dev/null
 
+readonly CONSTANT_TEAM_ID="<#teamId#>"
```
blyscuit, line 3: Small question, now `TEAM_ID` is a required field, but before we can use match without `TEAM_ID`?

ducbm051291: @blyscuit Yes, I think it is because the match uses automatic signing. Now we use update code signing to set code signing manually.

.github/workflows/test_upload_build_to_test_flight.yml
```diff
@@ -7,6 +7,7 @@ name: Test Upload Build to TestFlight
 ### API_KEY_ID
 ### ISSUER_ID
 ### APPSTORE_CONNECT_API_KEY
+### TEAM_ID
```
suho, line 10: We don't need to put `TEAM_ID` in secret. Lets add it in `Constant`

fastlane/Helpers/Match.swift
```diff
@@ -26,24 +26,72 @@ enum Match {
...
+    static func updateCodeSigning(type: MatchType, environment: Environment) {
+        // Update Code signing from automatic to manual
+        updateCodeSigningSettings(
+            path: Constant.projectPath,
+            useAutomaticSigning: .userDefined(false),
+            teamId: .userDefined(Constant.teamId),
+            targets: .userDefined([Constant.projectName]),
+            buildConfigurations: .userDefined(["\(type.buildConfiguration) \(environment.rawValue)"]),
+            codeSignIdentity: .userDefined(type.codeSignIdentity),
+            profileName: .userDefined("match \(type.method) \(environment.bundleId)")
```
suho, line 48: Please define the string for `profileName`, as we also use in `Build.swift`

.github/workflows/test_upload_build_to_test_flight.yml
ducbm051291: Ok, so I think I will revert this 😀

fastlane/Helpers/Match.swift
```diff
@@ -26,24 +26,72 @@ enum Match {
...
+            switch self {
+            case .development: return "Development"
+            case .adHoc: return "AdHoc"
+            case .appStore: return "AppStore"
+            }
+        }
+        
+        var buildConfiguration: String {
+            switch self {
+            case .development: return "Debug"
```
suho, line 86: Please double check, our build configurations:
- Debug Staging
- Release Staging
- Debug Production
- Release Production

ducbm051291: Please check because Ed's code is to combine MatchType build configuration with Environment raw.

`buildConfigurations: .userDefined(["\(type.buildConfiguration) \(environment.rawValue)"])`

set_up_test_testflight.sh
suho: @ducbm051291 using match is about setting code signing manually already. In this case, [`Team ID` is used when upgrading project](https://docs.fastlane.tools/actions/update_code_signing_settings/#:~:text=Team%20ID%2C%20is%20used%20when%20upgrading%20project). 

From [the example](https://docs.fastlane.tools/actions/update_code_signing_settings/#:~:text=%23%20manual%20code%20signing%0Aupdate_code_signing_settings(%0A%20%20use_automatic_signing%3A%20false%2C%0A%20%20path%3A%20%22demo%2Dproject/demo/demo.xcodeproj%22%0A)), we don't need to have `team_id`, so please recheck if we need it or not. If we need it, what is the error when running that command without team_id 🤔

suho:
@ducbm051291 btw, it seems you did wrong on setting up git user locally, please recheck it.
Screenshot 2023-04-21 at 09 43 51

ducbm051291:
Currently, we have the Environment enum, MatchType enum in Match and the Environment enum, BuildType enum in Constant. They're the same. I will merge them into one in Constant.

set_up_test_testflight.sh
ducbm051291: Without team id, we got this warning and the project couldn't be archived and failed with status code 65

CleanShot 2023-04-21 at 13 09 16@2x

suho: @ducbm051291 without `team_id` does not mean not setting team id. It seems you are still using `team_id` with `updateCodeSigningSettings`, please remove that parameter 🙈
```diff
updateCodeSigningSettings(
	path: Constant.projectPath,
	useAutomaticSigning: .userDefined(false),
-	teamId: .userDefined(Constant.teamId),
	targets: .userDefined([Constant.projectName]),
	buildConfigurations: .userDefined([Self.createBuildConfiguration(type: type, environment: environment)]),
	codeSignIdentity: .userDefined(type.codeSignIdentity),
	profileName: .userDefined(Self.createProfileName(type: type, environment: environment))
)
```

ducbm051291: @suho After removing setting teamId, the workflow failed to build app 😓

suho: what is the error? did you run on local?

ducbm051291: Same as usual, exit status code 65 

CleanShot 2023-04-21 at 14 22 07@2x

Here are the local runs

With set teamId

CleanShot 2023-04-21 at 14 32 56@2x

Without set teamId

CleanShot 2023-04-21 at 14 34 48@2x

suho: I see, so it seems `updateCodeSigningSettings` requires teamId. Thanks for the info

.github/workflows/test_upload_build_to_test_flight.yml
vnntsu: @ducbm051291 can we remove this comment code for `TEAM_ID`?

.github/workflows/test_upload_build_to_test_flight.yml
```diff
@@ -51,12 +52,13 @@ jobs:
         MATCH_REPO: ${{ secrets.MATCH_REPO }}
         API_KEY_ID: ${{ secrets.API_KEY_ID }}
         ISSUER_ID: ${{ secrets.ISSUER_ID }}
+        TEAM_ID: ${{ secrets.TEAM_ID }}
```
vnntsu, line 55: I think we don't get it from secrets anymore, can we remove it?

vnntsu (approved):
Please remove some redundant code, then we can merge this 🙏

.github/workflows/test_upload_build_to_test_flight.yml
ducbm051291: We still need it :D

.github/workflows/test_upload_build_to_test_flight.yml
ducbm051291: We still need it :D

PR #486, opened 2023-06-16
[#483] Add SwiftUI Option to the Template 

### LINKED ISSUE #483: Add SwiftUI Option to the Template
## Why

As part of #276.

User should be able to generate the project with SwiftUI + Combine as an option.

## Acceptance Criteria

- Include SwiftUI option to the template.
- SwiftUI option should not include UIKit source files.
- SwiftUI option should include a SwiftUI starting project.
- UIKit option should be available and unchanged. 

## Who Benefits?

Users.

blyscuit:
- Close #483

## What happened 👀

Add SwiftUI option to the template.

## Insight 📝

- Create folder to host code difference `tuist/Interfaces`
  - Add `sources/Application`
  - Add `sources/Data`
  - Add `sources/Presentation`
  - Add `Project/Podfile`.
  - Add `Project/.sourcery.yml`
- Create script to copy the file differences.
- SwiftUI needs an extra renaming for {Project}App.swift
- Reusable code stays in `{PROJECT}/sources`.
- Rename `setup_{script}` to `{script}_set_up`.
- Add Wiki

## Proof Of Work 📹

Screen Shot 2023-06-15 at 15 54 22
Screen Shot 2023-06-15 at 16 15 56


set_up_interface.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
suho, line 1: @blyscuit What do you think about using `swift` instead of bash?
```suggestion
#!/usr/bin/swift
```

blyscuit: Let me look in to that.

suho: This's optional

blyscuit: Will create a task for that. Although I don't think Swift will support replace all in files like `sed -I` so we will need to create some extension for it, which isn't ideal as it adds more failure points.

suho: @blyscuit Yes, that's a good point!. We could create POC first before having a task

Tuist/Interfaces/SwiftUI/Project/Podfile
```diff
@@ -0,0 +1,57 @@
+platform :ios, '14.0'
```
markgravity, line 1: Can we make this file reusable for both interfaces, we only need to update the iOS version, right? 🙏

blyscuit: The content are different, please not that we don't include RxSwift and UIKit related for SwiftUI.

Scripts/Swift/Extensions/FileManager+.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+extension FileManager {
```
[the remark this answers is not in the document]
blyscuit: What do you think about `FileManager+Extension.swift`?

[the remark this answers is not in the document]
markgravity: Sounds good! The `FileManager+Utils.swift` works for me as well 🙏

markgravity (approved):
LGTM

ducbm051291 (approved):
Based on my assessment, this pull request is good to go.

make.sh
```diff
@@ -188,13 +204,14 @@ echo "Remove script files and git/index"
...
 rm -f .git/index
+rm -rf Scripts
 git reset
 
 if [[ -z "${CI}" ]]; then
     rm -rf fastlane/Tests
     rm -f set_up_test_firebase.sh
     rm -f set_up_test_testflight.sh
-    sh deliverable_setup.sh
+    sh set_up_deliverable.sh
```
vnntsu, line 214: After completing this sh run, should we remove this `set_up_deliverable.sh`?
Reusing isn't valuable since we removed all stuff during setting up.

vnntsu (commented):
I have a concern that we should remove `ios-templates` workflows after generating, such as `test_install_script.yml`, ` test_upload_build_to_firebase.yml`, and `test_upload_build_to_testflight.yml` 💪

make.sh
blyscuit: `set_up_deliverable.sh` only opens the `Constants.swift` with Xcode. It is included so that user can modify `Constants.swift` without knowing where the file is located (deep within Fastlane folder).
The `make.sh` script prompt the user if they want to modify `Constants.swift` at project setup or later because some project might not have the necessary information for delivery yet.
We can keep this file for later use so we don't have to remove it.

vnntsu: As seeing, `set_up_deliverable.sh` contains two actions: 1) choosing CI/CD platform, 2) opening `Constants.swift`. It isn't very clear if the first action does nothing, even accidentally removing files from the pipeline in case users choose a different one.

So I prefer to separate action modifying `constants` to another script file. What do you think?

blyscuit: Sounds good. I will create an issue for that. For this PR let's keep it as is.

vnntsu (approved):
Please resolve the conflict 🙏

PR #496, opened 2023-07-07
[#490] Migrate `set_up_deliverable.sh` to Swift command

### LINKED ISSUE #490: Migrate `set_up_deliverable.sh` to Swift command
## Why

Migrate our setup commands to Swift command to ease maintenance for newcomers.

Currently `set_up_deliverable.sh` is doing two jobs as [pointed out by Su.t](https://github.com/nimblehq/ios-templates/pull/486#discussion_r1251451772). We should make the script simpler and remove the setup scripts as soon as it is success.
 
## Acceptance Criteria

- Create `SetUpCICDService.swift` for selecting a CICD platform.
  - Remove `SetUpCICDSerview.swift` if selected option is not `later`.
- Create `SetUpDeliveryConstants.swift` for modifying `Constants.swift`. 
- Run both scripts as part of `make.sh`.

## Who Benefits?

Template Developers

blyscuit:
- Close #490 

## What happened 👀

Migrate `set_up_deliverable.sh` to Swift Command.

## Insight 📝

- Create `SetUpCICDService.swift`.
- Create `SetUpDeliveryConstants.swift`.
- Added `SafeShell.swift` for running Bash in Swift.
- Currently `SetUpDeliveryConstants.swift` will be removed after `make.sh` because running `SetUpDeliveryConstants.swift` requires a long command. This will change when the Swift command is packaged.

## Proof Of Work 📹

Screen Shot 2023-07-07 at 12 06 07


Scripts/Swift/Helpers/SafeShell.swift
```diff
@@ -0,0 +1,20 @@
...
+    task.standardOutput = pipe
+    task.standardError = pipe
+    task.arguments = ["-c", command]
+    task.executableURL = URL(fileURLWithPath: "/bin/zsh")
+    task.standardInput = nil
+
+    try task.run()
+    
+    let data = pipe.fileHandleForReading.readDataToEndOfFile()
+    let output = String(data: data, encoding: .utf8)!
```
Shayokh144, line 17: can we avoid force unwrapping here and return a default string in case of nil?

blyscuit: Good suggestion. Let me try.

vnntsu (approved):

Scripts/Swift/SetUpCICDService.swift
```diff
@@ -0,0 +1,47 @@
...
+        case "b", "bitrise", "Bitrise":
+            self = .bitrise
+        case "c", "codemagic", "Codemagic":
+            self = .codemagic
+        default: self = .later
+        }
+    }
+}
+
+service = readLine() ?? service
```
Thieurom, line 27: I'd like to suggest two things 🙏 
- Instead of letting the user config the CICD service later as a `later` option, consider "Hit Enter if you want to edit/update later". We, developers, don't have to add a not-so-meaningful enum case. The users don't have to remember one more option–they might rerun this for the same or multiple projects, and easily to move forward.
- When user faces the `g`, `b`, and `c` options, they might select the unwanted option. For e.g, they want to choose `g` for GitHub but hit the wrong key on the keyboard, say `v` (a key nearby). In that case, the script process is as if the user selects `later`. It would confuse users. IMO we should try to ensure users select what they want, not providing a fallback. A simple and widely-used solution is to run a loop around `readLine()`.

Scripts/Swift/SetUpDeliveryConstants.swift
```diff
@@ -0,0 +1,18 @@
...
+
+print("Do you want to set up Constants values? (Can be edited later) [Y/n]: ")
+
+var arg = "y"
+
+arg = readLine() ?? arg
+
+switch arg.lowercased() {
+    case "y", "yes":
+        try safeShell("open -a Xcode fastlane/Constants/Constant.swift")
```
Thieurom, line 15: I don't try this yet, but the script might exit with a non-human-readable message (e.g., "EXIT with xxx code").

IMO, we should catch the error and exit gracefully with a helpful message like "Could not open Xcode. Please try again" 🙏

blyscuit: Good suggestion. Updated this part, this is the demo of the script failing. Screen Shot 2023-07-17 at 17 46 33

Scripts/Swift/SetUpCICDService.swift
blyscuit: Added while loop to this method. Here's the demo 
Screen Shot 2023-07-17 at 17 53 45
I didn't added the while loop in `DeliveryConstants` because that config has a lower significance than this config 🙏

Scripts/Swift/SetUpDeliveryConstants.swift
```diff
@@ -0,0 +1,20 @@
...
+
+print("Do you want to set up Constants values? (Can be edited later) [Y/n]: ")
+
+var arg = "y"
+
+arg = readLine() ?? arg
+
+switch arg.lowercased() {
+    case "y", "yes", "":
+        let error = try safeShell("open -a NotXcode fastlane/Constants/Constant.swift")
```
Thieurom, line 15: Please put it back to `Xcode` after testing 😅

blyscuit: 😆

Scripts/Swift/SetUpCICDService.swift
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/swift
+
+import Foundation
+
+let fileManager = FileManager.default
+
+enum CICDService {
+    case github, bitrise, codemagic, later, none
```
Thieurom, line 8: I prefer to remove the `later` (and `none`), but this is still good!

Scripts/Swift/SetUpDeliveryConstants.swift
```diff
@@ -0,0 +1,20 @@
...
+
+import Foundation
+
+let fileManager = FileManager.default
+
+print("Do you want to set up Constants values? (Can be edited later) [Y/n]: ")
+
+var arg = "y"
+
+arg = readLine() ?? arg
```
Thieurom, line 11: Could this be `let arg = readLine() ?? "y"`?

Thieurom (approved):

Scripts/Swift/SetUpCICDService.swift
blyscuit: Updated https://github.com/nimblehq/ios-templates/pull/496/commits/d5bf8197b64aabd8fe16808ca480c48a6dd4ba01

Scripts/Swift/SetUpDeliveryConstants.swift
blyscuit: Updated https://github.com/nimblehq/ios-templates/pull/496/commits/d5bf8197b64aabd8fe16808ca480c48a6dd4ba01

markgravity (approved):

Shayokh144 (approved):

PR #513, opened 2023-08-22
[#510] GitHub actions can remove custom keychain when needed

### LINKED ISSUE #510: GitHub Actions can remove custom Keychain when needed
## Why

As we are moving to `self-hosted-runner` for some of the projects, we would like to be able to manually remove Keychain for for the project when the project is done running.
 
## Who Benefits?

Self-hosted GH Action runners users.

## Acceptance Criteria

  - Create Keychain in Keychain default folder with name  `{Constants.keychainName}`
```
        if isCi() {
		// Create new custom self-hosted keychain
	} else {
	        // Use default keychain
        }
```
- Create a new fastlane lane `removeKeychain`.
    - Use fastlane [delete_keychain](https://docs.fastlane.tools/actions/delete_keychain/) if possible. 
    - Check for existing keychain with name `{Constants.keychainName}`.
    - Remove keychain with name `{Constants.keychainName}` if found.
```
        let isSelfHostedCiCd = EnvironmentParser.bool(key: "self_hosted_ci_cd")
        if isSelfHostedCiCd {
		// remove custom keychain
	}
```
- Modify variable `keychainName` in Constants.swift to allow automatic generation from `make.sh` with `{project_name}`.

- Modify GH Actions files to remove keychain if it was run with self-hosted variable `self_hosted_ci_cd` and feature. 

## Resources

https://www.notion.so/nimblehq/Instruction-Provision-Custom-MacOS-Runner-for-GitHub-Actions-ace7181d7bf24896804c7e3313c9c956?pvs=4

ducbm051291:
- Close #510 

## What happened 👀

- Add a new step: Remove Keychain in GitHub action workflows to remove the keychain after completing the job so in case the self-hosted runner, the system shall not show a dialog popup to ask for a keychain and freeze the job

- Modify variable keychainName in Constants.swift to allow automatic generation from make.sh with {project_name}.

## Insight 📝

- N/A

## Proof Of Work 📹

**Remove keychain step result**

CleanShot 2023-08-22 at 14 00 28@2x

**Modify variable keychainName in Constants.swift result**

CleanShot 2023-08-24 at 09 52 21

CleanShot 2023-08-24 at 09 55 40

### DIFF
--- a/.github/workflows/deploy_app_store.yml
+++ b/.github/workflows/deploy_app_store.yml
@@ -97,3 +97,8 @@ jobs:
           ${{ env.DSYM_OUTPUT_PATH }}
       env:
         TAG_TYPE: App_Store
+
+    - name: Remove keychain
+      if: ${{ always() }}
+      run: bundle exec fastlane removeKeychain
+      continue-on-error: true
--- a/fastlane/Constants/Constant.swift
+++ b/fastlane/Constants/Constant.swift
@@ -37,7 +37,6 @@ enum Constant {
     static let projectPath: String = "./\(projectName).xcodeproj"
     static let testOutputDirectoryPath = "./fastlane/test_output"
     static let infoPlistPath = "\(projectName)/Configurations/Plists/Info.plist"
-    static let keychainPath = "~/Library/Keychains/\(keychainName)-db"
 
     // MARK: Platform
 
--- a/fastlane/Constants/Constant.swift
+++ b/fastlane/Constants/Constant.swift
@@ -26,7 +26,7 @@ enum Constant {
     static let appleStagingTeamId = "<#teamId#>"
     static let appleProductionUserName = "<#userName#>"
     static let appleProductionTeamId = "<#teamId#>"
-    static let keychainName = "{PROJECT_NAME}"
+    static let keychainName = "{PROJECT_NAME}_keychain"
     static let matchURL = "git@github.com:{organization}/{repo}.git"
 
     // MARK: - Path
--- a/fastlane/Constants/Constant.swift
+++ b/fastlane/Constants/Constant.swift
@@ -37,6 +37,7 @@ enum Constant {
     static let projectPath: String = "./\(projectName).xcodeproj"
     static let testOutputDirectoryPath = "./fastlane/test_output"
     static let infoPlistPath = "\(projectName)/Configurations/Plists/Info.plist"
+    static let keychainPath = "~/Library/Keychains/\(keychainName)-db"
 
     // MARK: Platform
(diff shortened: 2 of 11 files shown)


blyscuit (commented):
Please add the keychain name auto populate.
```
Modify variable keychainName in Constants.swift to allow automatic generation from make.sh with {project_name}.
```

fastlane/Helpers/Keychain.swift
```diff
@@ -6,6 +6,8 @@
 //  Copyright © 2022 Nimble. All rights reserved.
 //
 
+import Foundation
+	
 enum Keychain {
 
     static func create() {
```
markgravity, line 13: Did we miss this change? 🤔 
```
if isCi() {
     // Create new custom self-hosted keychain
} else {
	// Use default keychain
}
```

ducbm051291: As we add a step to clean up the keychain, we no longer need this 👍

fastlane/Constants/Constant.swift
```diff
@@ -26,7 +26,7 @@ enum Constant {
     static let appleStagingTeamId = "<#teamId#>"
     static let appleProductionUserName = "<#userName#>"
     static let appleProductionTeamId = "<#teamId#>"
-    static let keychainName = "github_action_keychain"
+    static let keychainName = "{PROJECT_NAME}"
```
blyscuit, line 29: I suggest 
```suggestion
    static let keychainName = "{PROJECT_NAME}_keychain"
```

.github/workflows/deploy_app_store.yml
```diff
@@ -97,3 +97,8 @@ jobs:
           ${{ env.DSYM_OUTPUT_PATH }}
       env:
         TAG_TYPE: App_Store
+
+    - name: Remove keychain
+      if: ${{ always() }}
+      run: bundle exec fastlane removeKeychain
+      continue-on-error: true
```
Thieurom, line 104: Do we need `continue-on-error: true` here 🤔 That's because: 1, there's no subsequent step; 2. If the `removeKeychain` fails, the workflow still reports success for this step.

blyscuit: > If the removeKeychain fails, the workflow still reports success for this step

I think this is what we are aiming for because we can ignore the error in this step, if this step can fail then the whole CI will show failure which is also confusing. What do you think?

suho (approved):

blyscuit (approved):
Great work.

markgravity (approved):

PR #539, opened 2023-10-27
[#521] [Chore]  Update SwiftLint in GitHub Workflows to the newest version

### LINKED ISSUE #521: Update SwiftLint in GitHub Workflows to the newest version
## Why

Current version we are using does not support > Swift 5.1 and will fail linting.

## Acceptance Criteria

- Update `norio-nomura/action-swiftlint`
 
## Who Benefits?

Users.

ducbm051291:
- Close #521 

## What happened 👀

Update norio-nomura/action-swiftlint to latest version 3.2.1

## Insight 📝

- N/A

## Proof Of Work 📹

CleanShot 2023-10-27 at 16 59 27
CleanShot 2023-10-27 at 16 59 30
CleanShot 2023-10-27 at 16 59 33

### DIFF
--- a/.github/project_workflows/deploy_app_store.yml
+++ b/.github/project_workflows/deploy_app_store.yml
@@ -27,7 +27,7 @@ jobs:
         fetch-depth: 0
 
     - name: Run SwiftLint
-      uses: norio-nomura/action-swiftlint@3.2.1
+      uses: norio-nomura/action-swiftlint@master
       with:
         args: --strict
 
--- a/.github/project_workflows/deploy_app_store.yml
+++ b/.github/project_workflows/deploy_app_store.yml
@@ -27,7 +27,7 @@ jobs:
         fetch-depth: 0
 
     - name: Run SwiftLint
-      uses: norio-nomura/action-swiftlint@3.1.0
+      uses: norio-nomura/action-swiftlint@3.2.1
       with:
         args: --strict
 
--- a/.github/project_workflows/deploy_app_store.yml
+++ b/.github/project_workflows/deploy_app_store.yml
@@ -27,9 +27,9 @@ jobs:
         fetch-depth: 0
 
     - name: Run SwiftLint
-      uses: norio-nomura/action-swiftlint@master
+      uses: docker://norionomura/swiftlint:0.53.0_swift-5.7
       with:
-        args: --strict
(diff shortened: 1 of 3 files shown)


.github/project_workflows/deploy_app_store.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         fetch-depth: 0
 
     - name: Run SwiftLint
-      uses: norio-nomura/action-swiftlint@3.1.0
+      uses: norio-nomura/action-swiftlint@3.2.1
```
blyscuit, line 30: Improvement: Might need some investigation if this syntax works so we don't have to come up date this all the time https://stackoverflow.com/a/73259281/3736373.
```
      uses: norio-nomura/action-swiftlint@v3
```

ducbm051291: I updated to  `uses: norio-nomura/action-swiftlint@master` due to this github only has master branch 👍

blyscuit (approved):

.github/project_workflows/deploy_app_store.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         fetch-depth: 0
 
     - name: Run SwiftLint
-      uses: norio-nomura/action-swiftlint@3.1.0
+      uses: norio-nomura/action-swiftlint@master
```
Thieurom, line 30: TBH, I'd prefer to provide an explicit tag (i.e., `@3.2.1`) rather than `master`.

It might have breaking changes in the future (although a GitHub action is less likely, it still is a software 😅 ). Plus, we may not have to update as frequently as we think; we can ignore the patch updates (`3.2.x`), for example.

Thieurom (approved):

.github/project_workflows/deploy_app_store.yml
markgravity: How about using this solution, it works great on Smashburger and we can fix the version we want.
```
- name: Run SwiftLint
      uses: docker://norionomura/swiftlint:0.53.0_swift-5.7
      with:
        args: swiftlint --strict
```

blyscuit: No concern for me using Mark's approach.

markgravity (approved):

PR #544, opened 2023-11-17
[#519] Fix: Test run failed for not finding host application for staging scheme

### LINKED ISSUE #519: Fix: Test run failed for not finding host application for staging scheme
## Issue

Test run failed for not finding host application for staging scheme
 
## Expected

Our template used different product name for different configurations. For example:
DebugProduction contains: `PRODUCT_NAME = $(TARGET_NAME)`
DebugStaging contains: `PRODUCT_NAME = $(TARGET_NAME) Staging`
Sometime for getting different app name for different scheme we set `product name` in `build settings` using `PRODUCT_NAME` from configuration files like this:
Screenshot 2023-08-28 at 12 20 00 PM

As a result when we run `TARGET_NAME` scheme the app name in iPhone shows `TARGET_NAME` and when we run `TARGET_NAME staging` scheme the app name in iPhone shows `TARGET_NAME staging` as app name.

Screenshot 2023-08-28 at 12 24 02 PM

As a result when we run test for `AppName Staging` scheme, the test run throws an error because the product name and target name is different.

Screenshot 2023-08-28 at 12 19 25 PM

## Steps to reproduce

1. Create project from iOS Template
2. Change `product name` in build settings to use xcconfig files `PRODUCT_NAME`
3. Run test on `TARGET_NAME staging` scheme

Shayokh144:
- Close #519 

## What happened 👀

- Add new variable `APP_DISPLAY_NAME`  as `CFBundleName`
- Make `PRODUCT_NAME = $(TARGET_NAME)`

## Insight 📝

- `APP_DISPLAY_NAME` is used for showing different app name for different scheme
- `PRODUCT_NAME = $(TARGET_NAME)` will fix bug  `not finding host application
`
## Proof Of Work 📹

- App name

pow1

- All tests passed for both schemes

pow2


blyscuit (approved):

blyscuit:
> PRODUCT_NAME = $(TARGET_NAME) will fix bug  not finding host application 

This PR will resolve issue with host application because we will use `APP_DISPLAY_NAME` instead and `PRODUCT_NAME` now be the same for all scheme. Do I understand this correct? Or is the not finding host application still occur?

{PROJECT_NAME}/Configurations/XCConfigs/DebugStaging.xcconfig
```diff
@@ -7,7 +7,8 @@ ONLY_ACTIVE_ARCH = YES
 SWIFT_OPTIMIZATION_LEVEL = -Onone
 
 GCC_PREPROCESSOR_DEFINITIONS = $(inherited) DEBUG=1 STAGING=1
-PRODUCT_NAME = $(TARGET_NAME) Staging
+APP_DISPLAY_NAME = $(TARGET_NAME) Staging
+PRODUCT_NAME = $(TARGET_NAME)
```
Thieurom, line 11: We can safely remove all the `PRODUCT_NAME` entries in all `xcconfig` files.

It's because its value is always set to `$(TARGET_NAME)` by default. Also, whatever value is set here actually has no effect since, in the end, Xcode will resolve its value with the one provided at the target-level 🙏 (Meaning if we want to override, we have to override at the target-level, not at the config file level).

Thieurom (approved):

markgravity (approved):

PR #551, opened 2023-12-26
[#546] Support adding new devices directly from Github Action

### LINKED ISSUE #546: Add support to adding new devices directly from GitHub Actions
## Why

Currently, when we have new users who want to test the build on Firebase, we will need to add those devices to the Apple Developer Portal manually. It's better to create a new workflow on GitHub Actions to support that step so everyone can add new devices, and regenerate new profiles and builds.

## Acceptance Criteria

- Add new GitHub Actions workflow with `workflow_dispatch`
- Support inputs: devices, bundle ids, etc.

## Resources

https://github.com/nimblehq/match-creator/blob/main/.github/workflows/add_device_profile.yml

markgravity:
- Close #546 

## What happened 👀

- Add `add_device_profile.yml` workflow
- Add a new lane `addDevicesGenerateProfiles`

## Insight 📝

It's because we have a limitation on adding new device so I only test with a fake device information

## Proof Of Work 📹

CleanShot 2024-02-26 at 15 44 38

### DIFF
--- a/.github/project_workflows/add_device_profile.yml
+++ b/.github/project_workflows/add_device_profile.yml
@@ -1,0 +1,67 @@
+name: Test Add device and regenerate profiles with fastlane match
+
+# SECRETS needed:
+### SSH_PRIVATE_KEY for the match Repo
+### MATCH_PASSWORD
+### APPSTORE_CONNECT_API_KEY
+### API_KEY_ID
+### ISSUER_ID
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      devices:
+        description: 'JSON of devices. `{"Luka iPhone 6": "1234567890123456789012345678901234567890","Felix iPad Air 2": "abcdefghijklmnopqrstvuwxyzabcdefghijklmn"}`'
+        required: true
+        type: string
+      platform:
+        description: 'The platform'
+        required: true
+        default: 'ios'
+        type: choice
+        options:
+        - ios
+        - mac
+
+jobs:
+  create_files:
+    name: Create certificates and profiles
+    runs-on: macOS-latest
+    steps:
+    - name: Checkout the repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Install SSH key
+      uses: webfactory/ssh-agent@v0.5.4
+      with:
+        ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+
+    - name: Cache Pods
+      uses: actions/cache@v3
+      id: cocoapodCache
+      with:
+        path: Pods
+        key: ${{ runner.os }}-pods-${{ hashFiles('**/Podfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-pods-
+
+    - name: Bundle install
+      run: bundle install
+
+    - name: Add device and regenerate profiles with match
+      run: bundle exec fastlane addDevicesGenerateProfiles
+      env:
+        MATCH_PASSWORD: ${{ secrets.MATCH_PASS }}
+        DEVICES: ${{ inputs.devices }}
+        APPSTORE_CONNECT_API_KEY: ${{ secrets.APPSTORE_CONNECT_API_KEY }}
+        API_KEY_ID: ${{ secrets.API_KEY_ID }}
+        ISSUER_ID: ${{ secrets.ISSUER_ID }}
+        PLATFORM: ${{ inputs.platform }}
+
+    - name: Clean up keychain
+      if: ${{ always() }}
+      run: bundle exec fastlane remove_keychain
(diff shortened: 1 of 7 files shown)


.github/project_workflows/add_device_profile.yml
```diff
@@ -0,0 +1,67 @@
...
+        ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+
+    - name: Cache Pods
+      uses: actions/cache@v3
+      id: cocoapodCache
+      with:
+        path: Pods
+        key: ${{ runner.os }}-pods-${{ hashFiles('**/Podfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-pods-
```
phongvhd93, line 49: I don't think we need to cache the pod here.

fastlane/Constants/Constant.swift
```diff
@@ -28,12 +28,20 @@ enum Constant {
...
-        "issuer_id": Secret.appStoreIssuerIdKey,
-        "key": Secret.appstoreConnectAPIKey,
-        "in_house": false
-    ]
+    static let apiKey: [String: Any] = {
+        var key = Secret.appstoreConnectAPIKey
+        if let data = Data(base64Encoded: Secret.appstoreConnectAPIKey),
+           let decodedKey = String(data: data, encoding: .utf8) {
+            key = decodedKey
+        }
```
phongvhd93, line 36: Because the API key is stored in CI's secret, why do we need to encode it and decode it again? I think it is redundant to encode the key, it is not like the `.env` file.

fastlane/Fastfile.swift
```diff
@@ -176,6 +176,24 @@ class Fastfile: LaneFile {
         Match.syncCodeSigning(type: .adHoc, environment: .staging, isForce: true)
     }
 
+    func addDevicesGenerateProfilesLane() {
+        desc("Add device and regenerate profiles with match")
+
+        guard let data = Secret.devices.data(using: .utf8),
+              let devices = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] else {
+            return
```
phongvhd93, line 184: Can it throw some error to make the run fail?

ducbm051291: I don't think it's possible to throw any error because of `try?` here

fastlane/Constants/Constant.swift
markgravity: @phongvhd93 it's because we are using it as base64 🙏 

CleanShot 2024-07-03 at 10 27 30

fastlane/Fastfile.swift
markgravity: @phongvhd93 if we mark the func as `throws` the Fastlane won't recognize it's as a lane 🙏

.github/project_workflows/add_device_profile.yml
markgravity: @phongvhd93 updated at [c696076](https://github.com/nimblehq/ios-templates/pull/551/commits/c69607606f868ae1423df8160a43832e67bfb451)

phongvhd93 (approved):

fastlane/Fastfile.swift
```diff
@@ -176,6 +176,24 @@ class Fastfile: LaneFile {
         Match.syncCodeSigning(type: .adHoc, environment: .staging, isForce: true)
     }
 
+    func addDevicesGenerateProfilesLane() {
```
minhnimble, line 179: Shall we mention `Staging` since that is the env we only sync for now?
```suggestion
    func addDevicesGenerateProfilesForStagingLane() {
```

.github/project_workflows/add_device_profile.yml
```diff
@@ -0,0 +1,58 @@
...
+### APPSTORE_CONNECT_API_KEY
+### API_KEY_ID
+### ISSUER_ID
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      devices:
+        description: 'JSON of devices. `{"Luka iPhone 6": "1234567890123456789012345678901234567890","Felix iPad Air 2": "abcdefghijklmnopqrstvuwxyzabcdefghijklmn"}`'
```
minhnimble, line 15: Can shorten the JSON example 😄  
```suggestion
        description: 'JSON of devices. `{"Luka iPhone 6":"1234567890123456789012345678901234567890","Felix iPad Air 2":"abcdefghijklmnopqrstvuwxyzabcdefghijklmn"}`'
```

minhnimble (approved):
Rest LGTM!

fastlane/Fastfile.swift
markgravity: We may have the dev env as SM did, so it's better to keep a generic name here

Issue #560, opened 2024-07-18
Improve iOS Template Maker


markgravity:
## Why

It makes the user run iOS Template Maker easier by improving the print messages and input values...
 
## Who Benefits?

- Users

PR #561, opened 2024-07-18
[#560] [Chore] Improve iOS Template Maker

markgravity:
- Close #

## What happened 👀

- Add support functions to select, input or print a message to the Terminal: `ask() `, `confirm()`, `picker`, `step` & `write()`...
- Refactor `SetUp` classes to use new support functions

## Insight 📝

n/a

## Proof Of Work 📹

CleanShot 2024-07-17 at 16 27 04


Scripts/Swift/iOSTemplateMaker/Sources/iOSTemplateMaker/SetUpCICDService.swift
```diff
@@ -48,59 +61,62 @@ struct SetUpCICDService {
                 return nil
             }
         }
+
+        var title: String {
+            switch self {
+            case .macOSLatest:
+                "macos"
```
phongvhd93, line 68: the value here should be `macos-latest`. There are another options like `macos-14`

Scripts/Swift/iOSTemplateMaker/Sources/iOSTemplateMaker/Helpers/Terminal.swift
```diff
@@ -0,0 +1,127 @@
...
+        writeln(text.yellow)
+    }
+}
+
+func writeAt(_ row: Int, _ col: Int, _ text: String) {
+    moveTo(row, col)
+    write(text)
+}
+
+func ask(
```
phongvhd93, line 39: Does this trim the user's input? We might add spaces at the leading/trailing without notice.

Scripts/Swift/iOSTemplateMaker/Sources/iOSTemplateMaker/Helpers/Terminal.swift
```diff
@@ -0,0 +1,127 @@
...
+private func unexpand(text: String) -> String {
+    var result = ""
+    var spaceCount = 0
+    for char in text {
+        if char == " " {
+            spaceCount += 1
+        } else {
+            if spaceCount > 0 {
+                let tabs = String(repeating: "\t", count: spaceCount / 8)
+                let spaces = String(repeating: " ", count: spaceCount % 8)
```
phongvhd93, line 114: These two are duplicated in the lines 122 and 123. We can avoid this duplicate and allocate inside loops by declaring this outside the `for` loop.

Scripts/Swift/iOSTemplateMaker/Sources/iOSTemplateMaker/SetUpCICDService.swift
```diff
@@ -25,9 +25,22 @@ struct SetUpCICDService {
                 return nil
             }
         }
+
+        var title: String {
+            switch self {
+            case .github:
+                "Github"
```
phongvhd93, line 32: It should be `GitHub` 😁

Scripts/Swift/iOSTemplateMaker/Sources/iOSTemplateMaker/Helpers/Terminal.swift
markgravity: Do you you mean here? 🙏

phongvhd93: This function takes the user's input, right? I'm thinking about trimming users' input when they enter the app name and bundle ID...

markgravity: We won't need to handle it here, the `onValidate` is used from outside to validate the input.
For example:
It will show an error if we input `com.jfc.bundleid `
```
bundleIdProduction = ask(
    "Which is the bundle ID for the production environment?",
    note: "Ex: com.example.project",
    onValidate: validatePackageName
)
```

phongvhd93 (approved):

nkhanh44:
@markgravity I tried to run but got this error unexpectedly found nil, do you face this one too? 🤔 

Screenshot 2024-09-24 at 16 42 12

markgravity:
@nkhanh44 let me check 🙏

ducbm051291 (approved):

markgravity:
> @markgravity I tried to run but got this error unexpectedly found nil, do you face this one too? 🤔
> 
> Screenshot 2024-09-24 at 16 42 12

@nkhanh44 it works normally on my end, so I will need more context from your side to reproduce it 🙏 


CleanShot 2024-10-23 at 10 03 38

PR #619, opened 2026-04-01
[#590] Bump app version github action workflow

### LINKED ISSUE #590: Add bump app version Github Action workflow
## Why

To have a workflow that automatically bumps the app version and creates a pull request, which will speed up the common process at the beginning of the sprint. The workflow will have the following features:

- Manually trigger and ask the developer to input the next version of the application.
- Replace the current app version with the new one (usually the `MARKETING_VERSION`).
- Create a PR with type `Chore` and a corresponding title.
 
## Who Benefits?

iOS developers

phongvhd93:
- Close #590

## What happened 👀

- Allow the developer to select job to run in `create_release_pull_request_and_bump_version`
- Update the `Bump version` step to print the error
- Include the default MARKETING_VERSION in Project.swift

## Insight 📝

The current workflow `create_release_pull_request_and_bump_version` doesn't allow a developer to bump the app version alone, so I added the job selector to trigger a specific job in that workflow; all jobs will be selected by default.

## Proof Of Work 📹

Screenshot 2026-04-03 at 14 29 29

Screenshot 2026-04-03 at 14 43 45

## Summary by CodeRabbit

* **Configuration Updates**
  * Default app version set to 1.0.0 for projects and generated Xcode builds.

* **CI/CD Improvements**
  * Added a new workflow to create release pull requests on demand.
  * Hardened the version-bump workflow with stricter shell checks, validations, and safer MARKETING_VERSION updates.
  * Updated workflow naming and made PR base branch selection dynamic.


.cicdtemplate/.github/workflows/create_release_pull_request_and_bump_version.yml
```diff
@@ -62,8 +73,19 @@ jobs:
...
       - name: Bump version
         run: |
+          set -euo pipefail
           filename=$(find . -maxdepth 1 -name "*.xcodeproj" -exec basename {} .xcodeproj \; | head -n 1)
-          sed -i "" "s/MARKETING_VERSION = .*/MARKETING_VERSION = ${{ github.event.inputs.nextVersion }};/g" $filename.xcodeproj/project.pbxproj
+          pbxproj="$filename.xcodeproj/project.pbxproj"
+          if [[ -z "$filename" ]]; then
+            echo "Error: no .xcodeproj found in root directory"
+            exit 1
+          fi
```
dquangit: We should place the empty check before append `filename` with `.xcodeproj/project.pbxproj`. Otherwise, the empty check always returns false.

```suggestion
          if [[ -z "$filename" ]]; then
            echo "Error: no .xcodeproj found in root directory"
            exit 1
          fi
          pbxproj="$filename.xcodeproj/project.pbxproj"
```

dquangit (approved):
Looks good to me 👍 It works well, just took me a bit of time to verify 😅
Optional: I think we'd better separate into independent workflows instead of group them into 1

.cicdtemplate/.github/workflows/create_release_pull_request_and_bump_version.yml
```diff
@@ -3,6 +3,15 @@ name: Create the Release pull request and Bump the next version
...
+        required: false
+        default: 'all'
+        type: choice
+        options:
+          - all
+          - bumpVersion
+          - release
       nextVersion:
         description: "Next version (eg. 1.0.0)"
         required: true
```
thinh2k1310, line 8: Optional: 
`nextVersion` still required when only `release` runs. Consider making it not required and adding a small guard in the `bump` job to validate `nextVersion`.

thinh2k1310 (approved):

.cicdtemplate/.github/workflows/create_release_pull_request_and_bump_version.yml
phongvhd93: I split the workflow, it is annoying when the workflow keeps asking for the next version while the developer only wants to make a release.

PR #684, opened 2026-04-13
[#635] Add authenticated network interceptor and token refresh handling

### LINKED ISSUE #635: Add authenticated network interceptor and token refresh handling
## Why

Authenticated requests, token injection, and `401` recovery are recurring needs in almost every product app.

This is higher leverage when provided once in the template than reimplemented per project.

## Acceptance Criteria

- The network layer can attach access tokens to authenticated requests.
- `401` responses trigger a refresh flow and retry behavior when appropriate.
- Refresh coordination avoids duplicated refresh calls during concurrent request failures.
- If refresh fails, the session is cleared, and the app can transition to a signed-out state.

## Implementation Notes

- Build on top of the existing `NetworkAPI` layer in `template/Modules/Data/Sources/NetworkAPI/`.
- Prefer Swift Concurrency with `actor`-based coordination for refresh state.
- Keep request interception generic so it can work with different auth providers later.

thinh2k1310:
- Close #635 

## What happened 👀

- `TokenRefreshCoordinator`: Ensures concurrent callers share a single in-flight refresh (queue + resume), with retries and backoff, and maps failures to `APIAuthenticationError`.
- `AuthenticationInterceptor`: 
  - `adapt`: injects `Authorization` using the coordinator’s validAccessToken(). 
  - `retry`: on 401, calls `refresh()` and retries the request once.
- RefreshTokenClient: Abstraction for the refresh HTTP call (default uses existing auth request configuration + TokenSet decoding).
- `APIAuthenticationError`: Typed errors for missing token, expired refresh, and refresh failure.
- `NetworkAPIProtocol`: Adds `performAuthenticatedRequest`; default extension keeps backward compatibility by delegating to `performRequest`.
- `NetworkAPI`: Holds two Session instances: one for unauthenticated performRequest, one for performAuthenticatedRequest (with interceptor when provided). Reuses sessions instead of creating a new `Session()` per call.
- Registers `tokenRefreshCoordinator`, `authenticationInterceptor`, and `networkAPI` as singletons so all traffic shares one coordinator/interceptor pair.
- Tests: Removes `NetworkAPISpec` (Quick/Nimble), adds `NetworkAPITests` and `TokenRefreshCoordinatorTests`.

## Insight 📝

Alternatives considered, but I did not use here:
- New coordinator/interceptor per resolution (non-singleton DI) - Would break coordination: each actor would have its own isRefreshing / waiters, so parallel refreshes could still occur.
- Singleton only on `NetworkAPI` with coordinator owned privately - Possible, but separate Factory registrations make testing overrides and explicit lifecycle easier without coupling everything to NetworkAPI’s initializer.

## Proof Of Work 📹

- Tested on GT-iOS app

https://github.com/user-attachments/assets/e8d44eb3-2193-4cda-be58-87775a03504e

- Tests run successfully
Screenshot 2026-04-13 at 18 04 14
Screenshot 2026-04-13 at 18 04 22

## Summary by CodeRabbit

* **New Features**
  * Automatic authentication interceptor with token refresh and explicit authentication error types
  * Centralized token refresh coordinator with proactive refresh and retry/backoff behavior
  * Support for authenticated API requests using a dedicated authenticated session

* **Refactor**
  * Network client now reuses sessions (authenticated vs unauthenticated) instead of creating per-request

* **Tests**
  * Added comprehensive tests for token refresh coordination, interceptor behavior, concurrency and retry scenarios
  * Removed an older/superseded NetworkAPI test suite


ducbm051291 (commented):
Nice work. Could we also add a focused unit test suite for the AuthenticationInterceptor to directly cover the adapt/retry behavior? In particular: 
- adapt injects Authorization: Bearer <token>
- retry triggers only on first 401 and returns .retry after successful refresh
- refresh failure (or non-401 / retryCount >= 1) returns doNotRetryWithError.

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    private let sessionRepository: any SessionRepositoryProtocol
+    private let refreshClient: any RefreshTokenClient
+
+    private var isRefreshing = false
+    private var pendingContinuations: [CheckedContinuation<Void, any Error>] = []
+
+    public init(
+        sessionRepository: any SessionRepositoryProtocol,
+        refreshClient: any RefreshTokenClient,
```
phongvhd93, line 51: Why can't we use the same network client but create a separate RefreshTokenClient?

phongvhd93 (commented):
We need to test the locking mechanism of the refresh token coordinator.  The scenario

- GET /me -> got 401 -> GET /addresses while refreshing token

Expectation

- GET /addresses will be added to the queue and automatically resume after the refresh token is done.
- Automatically remove all the tokens when the coordinator hits the refreshing limit, and get the correct amount of delay between each refresh token call.

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+        isRefreshing = true
+
+        do {
+            try await performRefresh()
+            resumeAll(throwing: nil)
+        } catch {
+            resumeAll(throwing: error)
+            throw error
+        }
+    }
```
Thieurom, line 99: Maybe we can replace the `isRefreshing` and `pendingContinuations` with a `Task`, the implementation could be simpler?
```swift
private var refreshTokenTask: Task<TokenSetProtocol, Error>?

public func validAccessToken() async throws -> String {
    if let refreshTokenTask {
        return try await refreshTokenTask.value.accessToken
    }

    if let tokenSet = await sessionRepository.currentTokenSet() {
        return tokenSet.accessToken
    }
    throw APIAuthenticationError.missingToken
}
```

```swift
private func performRefresh() async throws -> TokenSetProtocol {
    // 
}

public func refresh() async throws -> TokenSetProtocol {
    if let refreshTokenTask {
        return try await refreshTokenTask.value.accessToken
    }

    let task = Task<TokenSetProtocol, Error> {
        do {
            let newTokenSet = try await performRefresh()
            return newTokenSet
        } catch {
            
        }
    }
    refreshTokenTask = task
    defer { refreshTokenTask = nil }
    return try await task.value
}
```

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+                pendingContinuations.append(continuation)
+            }
+        }
+
+        guard let tokenSet = await sessionRepository.currentTokenSet() else {
+            throw APIAuthenticationError.missingToken
+        }
+
+        return tokenSet.accessToken
+    }
```
Thieurom, line 80: Should we proactively refreshes the token before it expires to avoid mid-request 401 failures?

Ref: https://github.com/Alamofire/Alamofire/blob/f73a2fcb60198ef2b92dc3b6074b18f98ccee875/Source/Features/AuthenticationInterceptor.swift#L33

1 minute threshold could be a reasonable value I think.

Thieurom (commented):
Nice implementation!

Thieurom:
> We need to test the locking mechanism of the refresh token coordinator.

`TokenRefreshCoordinator` is implemented as `actor`, I think it would work. However, we should cover it with unit tests.

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinator.swift
thinh2k1310: If we reuse NetworkAPI, it will create a circular dependency:
NetworkAPI → AuthenticationInterceptor → TokenRefreshCoordinator → NetworkAPI
I think the current implementation is a cleaner choice:
NetworkAPI → AuthenticationInterceptor → TokenRefreshCoordinator → RefreshTokenClient
Moreover, we can easily create a mock for `RefreshTokenClient` for testing purposes. 🙏

Thieurom (approved):

thinh2k1310:
Thank @ducbm051291, @phongvhd93, @Thieurom, all your comments resolved at [49ef2bf](https://github.com/nimblehq/ios-templates/pull/684/commits/49ef2bf7c72f8693158cf35655672639d161c0ac). Also update the POW 😄

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinator.swift
```diff
@@ -0,0 +1,155 @@
...
+//
+
+import Alamofire
+import Domain
+import Model
+import Foundation
+
+/// Configuration for token refresh retry behavior.
+public struct TokenRefreshConfiguration {
+    /// Maximum number of retry attempts for failed refresh operations.
```
ducbm051291, line 13:
```suggestion

    /// Maximum number of retry attempts for failed refresh operations.
```

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinatorProtocol.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  TokenRefreshCoordinatorProtocol.swift
+//
+
+/// Describes the coordinator responsible for obtaining a valid access token and refreshing it.
+public protocol TokenRefreshCoordinatorProtocol: Actor {
+    /// Returns the current valid access token.
```
ducbm051291, line 8:
```suggestion

    /// Returns the current valid access token.
```

template/Modules/Data/Sources/NetworkAPI/Authentication/TokenRefreshCoordinatorProtocol.swift
```diff
@@ -0,0 +1,15 @@
...
+    ///
+    /// - Throws: `APIAuthenticationError.missingToken` if no token set is available
+    func validAccessToken() async throws -> String
+    /// Performs a token refresh, coordinating concurrent callers so only one in-flight refresh occurs.
+    ///
+    /// - Throws: `APIAuthenticationError` or the underlying network error on failure
+    func refresh() async throws
+}
```
ducbm051291:
```suggestion
}

```

template/Modules/Data/Tests/Sources/Specs/NetworkAPI/AuthenticationInterceptorTests.swift
```diff
@@ -0,0 +1,385 @@
...
+        )
+    }
+
+    func clearSession() throws {
+        token = nil
+    }
+}
+
+private actor MockRefreshClient: RefreshTokenClient {
+    enum Outcome {
```
ducbm051291, line 324:
```suggestion

    enum Outcome {
```

template/Modules/Data/Tests/Sources/Specs/NetworkAPI/AuthenticationInterceptorTests.swift
```diff
@@ -0,0 +1,385 @@
...
+        case .success(let token):
+            return token
+        case .failure(let error):
+            throw error
+        }
+    }
+}
+
+private actor MockTokenRefreshCoordinator: TokenRefreshCoordinatorProtocol {
+    private let sessionRepository: any SessionRepositoryProtocol
```
ducbm051291, line 348:
```suggestion

    private let sessionRepository: any SessionRepositoryProtocol
```

template/Modules/Data/Tests/Sources/Specs/NetworkAPI/TokenRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,375 @@
...
+
+    func clearSession() throws {
+        token = nil
+    }
+}
+
+private actor CountingRefreshClient: RefreshTokenClient {
+
+    enum Outcome {
+        case success(TokenSet)
```
ducbm051291, line 282:
```suggestion

        case success(TokenSet)
```

template/Modules/Data/Tests/Sources/Specs/NetworkAPI/TokenRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,375 @@
...
+
+    func performRefreshCount() -> Int {
+        count
+    }
+}
+
+private actor DelayedRefreshClient: RefreshTokenClient {
+
+    enum Outcome {
+        case success(TokenSet)
```
ducbm051291, line 312:
```suggestion

        case success(TokenSet)
```

template/Modules/Data/Tests/Sources/Specs/NetworkAPI/TokenRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,375 @@
...
+
+    func performRefreshCount() -> Int {
+        count
+    }
+}
+
+private actor TimingRefreshClient: RefreshTokenClient {
+
+    enum Outcome {
+        case success(TokenSet)
```
ducbm051291, line 346:
```suggestion

        case success(TokenSet)
```

ducbm051291 (approved):
LGTM. Just some minor format issues

PR #686, opened 2026-04-13
[#685] Reduce GitHub Actions runtime and cost

### LINKED ISSUE #685: Reduce GitHub Actions runtime and cost
## Why

Our GitHub Actions workflows are using more runtime than necessary, especially on macOS jobs. Several workflows run on broad triggers even when the change only affects docs or wiki content, some jobs fetch full git history when only the latest commit is needed, and a few cache paths do not target the actual SwiftPM or Tuist build folders. There is also duplication across similar Firebase and TestFlight workflows, which makes the CI surface larger than necessary and increases maintenance costs.

This chore will tighten workflow triggers, use the smallest checkout depth that still preserves release behavior, fix cache placement so restores are effective, and reduce duplicated setup where possible. The goal is to cut GitHub-hosted runner minutes without changing release behavior.

## Who Benefits?

Everyone working on the template repo benefits from faster feedback and lower CI cost. Maintainers benefit from less duplicated workflow logic and easier upkeep. The iOS chapter and future generated-project teams benefit from a leaner, more reliable CI/CD baseline.

ducbm051291:
Closes #685

## What happened 👀
- Added path filters so docs-only changes do not trigger expensive CI jobs.
- Cached `scripts/iOSTemplateMaker/.build` to reuse generator build artifacts.
- Reduced checkout depth where full history is not needed.

## Insight 📝

This lowers GitHub Actions runtime and cost without changing release behavior. I kept `cleanup_cache` on the self-hosted Mac runner because that workflow belongs there.

## Proof Of Work 📹

- YAML validated locally
- `git diff --check` passed

## Summary by CodeRabbit

* **Chores**
  * Optimized CI/CD workflows to improve build efficiency and reduce unnecessary workflow executions for documentation-only changes.

### DIFF
--- a/.github/workflows/cleanup_cache.yml
+++ b/.github/workflows/cleanup_cache.yml
@@ -4,15 +4,14 @@ on:
     types:
       - closed
 
+permissions:
+  actions: write
+  contents: read
+
 jobs:
   cleanup_cache:
-    runs-on: macOS-latest
+    runs-on: ubuntu-latest
     steps:
-      - name: Check out code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # actions/checkout@v6.0.2
-        with:
-          fetch-depth: 1
-
       - name: Cleanup
         run: |
           gh extension install actions/gh-actions-cache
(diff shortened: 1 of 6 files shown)


thinh2k1310 (approved):

phongvhd93 (approved):
I think we can apply the fech-depth: 1 to these workflows.

